PR #2, opened 2016-10-13
Some bugfixes

JadonKrys:


file_catalog/server.py
```diff
@@ -227,16 +229,52 @@ def get(self):
     @coroutine
     def post(self):
         metadata = json_decode(self.request.body)
-        ret = yield self.db.get_file({'file_name':metadata['file_name']})
+
+        # check metadata for mandatory fields
+        if not all(x in metadata for x in ['locations', 'uid', 'checksum']):
```
dsschult: I think this is better written as
```python
if not set(('uid','checksum','locations')).issubset(metadata):
```

But if you don't like the set syntax, then just convert the list to a tuple in the current line.

file_catalog/server.py
```diff
@@ -227,16 +229,52 @@ def get(self):
...
+            # locations (url) is mandatory
+            self.send_error(400, message='mandatory metadata missing',
+                            file=self.files_url)
+            return
+        elif any(x in metadata for x in ['_id', 'mongo_id']):
```
dsschult: Same as previous condition

file_catalog/server.py
```diff
@@ -227,16 +229,52 @@ def get(self):
...
+            # forbidden fields
+            self.send_error(400, message='forbidden attributes',
+                            file=self.files_url)
+            return
+        elif not isinstance(metadata['locations'], list):
+            # locations needs to be a list
+            self.send_error(400, message='member `locations` must be a list',
+                            file=self.files_url)
+            return
+        elif not len(metadata['locations']):
```
dsschult: This is probably better as

```python
elif not metadata['locations']:
```

No need to get the length.

file_catalog/server.py
```diff
@@ -227,16 +229,52 @@ def get(self):
...
+            # location needs have at least one entry
+            self.send_error(400, message='member `locations` must be a list with at least one url',
+                            file=self.files_url)
+            return
+        elif any(not len(l) for l in metadata['locations']):
```
dsschult: Similar to above, better as:

```python
elif not all(l for l in metadata['locations'])
```

Actually, now that I think about it you could remove the previous check.  This one would catch that condition as well.

JadonKrys: I don't think so since `all([])` is `True`

dsschult: Odd, `any([])` is `False`, but `all([])` is `True`.  I thought they'd match.

Issue #10, opened 2016-10-25
Put any configurations into a config file


JadonKrys:
Use `ConfigParser`.

Things that should come into the config file:

- List of mandatory fields
- List of forbidden fields
- Maximum number of results when querying the file list (see also #8)
- Port of server

Anything else?

Issue #9, opened 2016-10-24
Server should take care about the field `metadata_modify_date`


JadonKrys:
Server needs to add/update `metadata_modify_date` automatically. This field shouldn't be allowed to be set by the user.

PR #12, opened 2016-10-28
Closes #8, #9, and #10 / added proper server response for invalid `mongo_id`

### LINKED ISSUE #8: Add maximum number of results when querying the file list
It is currently not any more in the code.

I would suggest adding a config file where one can specify such parameters in order to avoid the need to modify the actual code.

JadonKrys:


file_catalog/__main__.py
```diff
@@ -12,6 +14,22 @@ def main():
     parser.add_argument('--debug', action='store_true', default=False, help='Debug flag')
     args = parser.parse_args()
     kwargs = {k:v for k,v in vars(args).items() if v}
+
+    # create config dict
+    config = Config(os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'server.cfg')))
```
dsschult: The path assumption here seems dangerous. Probably better to pass it in as an argument to the parser.

file_catalog/config.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, path):
+        self.path = path
+        self.list_cache = {}
+
+        # read file
+        tmp = SafeConfigParser()
+        tmp.read(path)
+        self.__config_options_dict(tmp)
+
+    def __config_options_dict(self, config):
```
dsschult: Prefer a single leading underscore for non-public methods:
https://www.python.org/dev/peps/pep-0008/#id45

file_catalog/config.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Parses the value of a given `section`/`name` pair and returns a list.
+        It is expected that the value is a comma separated list, e.g. `a, b, c,d,e ,f , g`.
+        Note that all white spaces are removed before and after a comma.
+        """
+        if section in self.list_cache:
+            if name in self.list_cache[section]:
+                return self.list_cache[section][name]
+        else:
+            self.list_cache[section] = {}
```
dsschult: How many times are we actually executing this function?  If not often, I'd advise to just remove the cache.

Also, this is a handy decorator if you really want to use a cache:
https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize
Works even better in python3 with `functools.lru_cache`

file_catalog/server.py
```diff
@@ -63,17 +67,25 @@ def sort_dict(d):
 
     return od
 
+def set_last_modification_date(d):
+    d['meta_modify_date'] = str(datetime.datetime.now())
```
dsschult: I prefer using the utc date everywhere possible, just to avoid time zones.

file_catalog/server.py
```diff
@@ -292,28 +317,35 @@ def post(self):
...
+                ret['_links'] = {
+                    'self': {'href': os.path.join(self.files_url,mongo_id)},
+                    'parent': {'href': self.files_url},
+                }
+    
+                self.write(ret)
+            else:
+                self.send_error(404, message='not found')
+        except pymongo.errors.InvalidId:
+            self.send_error(400, message='`%s` is not a valid mongo_id, it must be a 12-byte input or a 24-character hex string' % mongo_id)
```
dsschult: Don't echo input back to the user. This is actually a security corner case that probably doesn't matter to us, but is good policy.  So, just say "not a valid mongo_id".

Same goes for all subsequent errors.

file_catalog/validation.py
```diff
@@ -5,56 +5,56 @@ def is_valid_sha512(hash_str):
...
-def has_forbidden_attributes_creation(apihandler, metadata):
+def has_forbidden_attributes_creation(apihandler, config, metadata):
     """
     Checks if dict (`metadata`) has forbidden attributes.
 
     Returns `True` if it has forbidden attributes.
     """
 
-    if set(('_id', 'mongo_id')) & set(metadata):
+    if set(config.get_list('metadata', 'forbidden_fields_creation')) & set(metadata):
```
dsschult: Because of the fact that the config can't change while we're running, I might be inclined to package up the validation methods into a class, which takes the config as input.  That way it can store these lists locally instead of directly referencing the config each time.

file_catalog/server.py
JadonKrys: Good point, didn't think about

PR #38, opened 2019-08-13
implement configuration of mongodb authentication

vbrik:


file_catalog/mongo.py
```diff
@@ -18,15 +19,14 @@
...
-        if port:
-            kwargs['port'] = port
-        logger.info('MongoClient args: %r', kwargs)
-
-        self.client = MongoClient(**kwargs).file_catalog
+    def __init__(self, host=None, port=None, username=None, password=None):
+        password_hash = (hashlib.sha256(password.encode('utf-8')).hexdigest() if password else None)
+        logger.info('MongoClient args: host=%s, port=%s, username=%s, '
+                               'sha256(password)=%s' %
+                            (host, port, username, password_hash))
```
dsschult: What's the reason for the password hash?  Is it just logging?  If so, that seems unnecessary.

file_catalog/server.py
```diff
@@ -139,6 +135,18 @@ def __init__(self, config, port=8888, db_host='localhost', db_port=27017, debug=
         )
         app.listen(port)
 
+    def print_config(self, config, port, debug, db_host, db_port, db_user, db_pass):
+        logger.info('db host: %s', db_host)
+        logger.info('db port: %s', db_port)
+        logger.info('db user: %s', db_user)
+        db_pass_hash = (hashlib.sha256(db_pass.encode('utf-8')).hexdigest() if db_pass else None)
+        logger.info('db pass hash: %s', db_pass_hash)
```
dsschult: Again, I probably wouldn't even log the hash here.

vbrik: The idea is not to print/log clear-text password, but still make it possible to verify that the correct password is used.

file_catalog/server.py
```diff
@@ -83,20 +82,23 @@ def set_last_modification_date(d):
...
             raise Exception('bad template path')
 
-        # print configuration
         logger.info('db host: %s', db_host)
         logger.info('db port: %s', db_port)
+        logger.info('db user: %s', db_user)
         logger.info('server port: %r', port)
         logger.info('debug: %r', debug)
-        logger.info('config: %r', config)
+        redacted_config = dict(config)
```
blinkdog: Not really necessary in this specific case, but consider:
`redacted_config = copy.deepcopy(config)`

PR #40, opened 2019-08-15
make it possible to configure token signing algorithm

vbrik:


tests/test_collections.py
```diff
@@ -10,8 +11,9 @@
 
 class TestCollectionsAPI(TestServerAPI):
     def test_10_collections(self):
-        self.start_server()
         token = self.get_token()
+        alg = jwt.get_unverified_header(token)['alg']
+        self.start_server(config_override={'TOKEN_ALGORITHM':alg})
```
dsschult: Instead of analyzing the token for the algorithm, it's probably easier in the tests to just specify the TOKEN_ALGORITHM in the .circleci config.  Is there a reason not to do that?

vbrik: I just didn't think about that. Will change

PR #43, opened 2019-09-16
Requirements refresh and Vagrantfile

blinkdog:
- Updated the versions for requirements in `requirements.txt`
- Added a `Vagrantfile` for spinning up File Catalog VMs
- Updated `.gitignore` to the latest one supplied by GitHub for Python project (+ an ignore for `.vagrant` in case someone makes use of the `Vagrantfile`)
- Added some documentation to `README.md`


.gitignore
```diff
@@ -75,19 +78,30 @@ target/
...
 
 # Spyder project settings
 .spyderproject
+.spyproject
 
 # Rope project settings
 .ropeproject
 
-*.swp
-*.swo
```
vbrik: Do you want to get rid of these? They are vi's swap files

blinkdog: Good catch. I think we want to keep swap files out of version control.
I've replaced it in the latest commit.

PR #45, opened 2019-10-14
First pass at adding locations via POST

blinkdog:
JIRA: https://wipacit.atlassian.net/browse/DEV-721

This is a first pass at adding a handler for `POST /api/files/UUID/locations` to allow atomic addition of new locations to a file catalog record.

No unit tests yet, this PR is offered for code review.

### DIFF
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -199,7 +199,7 @@ class Mongo(object):
         """Append distinct elements to arrays within a file document."""
         # build the query to update the file document
         update_query = {"$addToSet": {}}
-        for key in metadata.keys():
+        for key in metadata:
             if isinstance(metadata[key], list):
                 update_query["$addToSet"][key] = {"$each": metadata[key]}
             else:
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import datetime
 import logging
 try:
     from collections.abc import Iterable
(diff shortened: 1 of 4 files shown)


blinkdog:
I've written it to do multiple, so I've changed the comment at the top. Honestly, I don't think our use-case goes beyond adding a single location at a time, but the `locations` field is always an array. I thought it best to keep that convention in the interface.

And eek, this rabbit hole goes a little deeper than I expected. I thought the `self.db.update_file` call was doing some kind of [optimistic locking](https://sazzer.github.io/blog/2015/07/18/Optimistic-Locking-in-MongoDB/) to avoid the race conditions that we talked about adding the POST route to avoid.

However, the PATCH route basically does the same read/modify/(over)write operation that LTA would be doing, just on the server instead of the client. Looking at the `update_file` method in the `Mongo` class, the `self.client.files.update_one` method from pymongo is sync, but `@run_on_executor` allows it to run in parallel:

```python
    @run_on_executor
    def update_file(self, uuid, metadata):
        result = self.client.files.update_one({'uuid': uuid},
                                              {'$set': metadata})

        if result.modified_count is None:
            logger.warn('Cannot determine if document has been modified since `result.modified_count` has the value `None`. `result.matched_count` is %s' % result.matched_count)
        elif result.modified_count != 1:
            logger.warn('updated %s files with id %r',
                        result.modified_count, metadata_id)
            raise Exception('did not update')
```

It's probably worth drilling down a layer to use MongoDB's array update operators [$addToSet, $each](https://docs.mongodb.com/manual/reference/operator/update/addToSet/#addtoset-modifiers).

However, implementing the optimistic locking requires a little bit more thought, and then large modifications to make usage consistent, in order to be truly useful. That's beyond the scope of what I'd like to accomplish with this PR.

dsschult:
At some point it would be worth overhauling the code to make use of async mongo calls (they didn't exist when we started this a few years ago).  And also editing other server methods to do proper atomic operations.

As for optimistic locking, you've just used a bad word there.  I tried that sort of scheme in IceProd, and I removed it in a refactoring because it was causing all sorts of problems.  Simpler is better, unless you really, really need the complexity.

For the scope of this PR, I think it's useful to make a new `Mongo` method to handle the atomic update to a file using an array update.  This is sort-of a special case, where we don't just want to set the entire field.

file_catalog/mongo.py
```diff
@@ -192,3 +193,25 @@ def create_snapshot(self, metadata):
     @run_on_executor
     def get_snapshot(self, filters):
         return self.client.snapshots.find_one(filters, {'_id':False})
+
+    @run_on_executor
+    def append_distinct_elements_to_file(self, uuid, metadata):
+        """Append distinct elements to arrays within a file document."""
+        # build the query to update the file document
+        update_query = {"$addToSet": {}}
+        for key in metadata.keys():
```
dsschult: Suggest `for key in metadata`.  There shouldn't be a need to call `.keys()`.

blinkdog: Ahhh yes, good call. Should be fixed now.

dsschult:
Also note, if you need the document back, you could use
https://docs.mongodb.com/manual/reference/method/db.collection.findOneAndUpdate/

blinkdog:
If there are no objections, I think this one is ready to merge.

PR #46, opened 2020-01-17
Metadata updates

ric-evans:
I expanded metadata collection, and added scripts for running with Condor. L2 years 2007-2020 are supported, more metadata is available in later years.

### DIFF
--- a/resources/indexer.py
+++ b/resources/indexer.py
@@ -166,7 +166,7 @@ def _parse_gaps_dict(gaps_dict):
     return gaps, livetime, first_event_dict, last_event_dict
 
 
-def _get_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gcd_files=""):
+def _get_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gcd_file=""):
     """Return metadata for one file"""
     path = file.path
     start_dt, end_dt, create_date, software = _parse_xml(path, run_meta_xml)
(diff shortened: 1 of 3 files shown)


resources/indexer_env.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+eval `/cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/setup.sh`
+. /scratch/eevans/env-fc-indexer/bin/activate
+/cvmfs/icecube.opensciencegrid.org/py3-v4.1.0/RHEL_7_x86_64/metaprojects/combo/stable/env-shell.sh
```
dsschult: Can do this instead:
`$SROOT/metaprojects/combo/stable/env-shell.sh`

resources/indexer.py
```diff
@@ -1,13 +1,227 @@
...
+    '2015': 'IC86-5',
+    '2016': 'IC86-6',
+    '2017': 'IC86-7',
+    '2018': 'IC86-8',
+    '2019': 'IC86-9',
+    '2020': 'IC86-10'
+}
+
+
+def _get_year_from_season(season):
```
dsschult: Is there a reason for the leading underscore on function names? Just wondering.

resources/indexer.py
```diff
@@ -1,13 +1,227 @@
...
 from time import sleep
 
+import requests
+import xmltodict
+import yaml
+from icecube import dataclasses, dataio
+from rest_tools.client import RestClient
+
+SEASONS = {
+    '2005': 'ICstring9',
```
dsschult: Not sure if the proper name for 2005 should be "IC1".  Not like it really matters, since no one uses that data anymore.

resources/indexer.py
ric-evans: I use that to denote private functions. In this case there isn't much of a difference between public and private, but it's a habit.

resources/indexer.py
ric-evans: I'm using what Kevin gave me, but I'm not reading data that early anyways.

ric-evans:
updates to indexer, for L2 directories/files

PR #51, opened 2020-02-06
add an overall test for mongodb unindexed queries

dsschult:
Take a wrapper approach, and flag any unindexed mongodb queries in the whole test suite.


dsschult:
That was the idea, once it was reviewed.

PR #52, opened 2020-02-06
PFFilt metadata

ric-evans:
Refactored a lot but the same logic is still there. Essentially class inheritance has replaced if-blocks for processing levels. Added support for PFFilt files.

### DIFF
--- a/resources/indexer.py
+++ b/resources/indexer.py
@@ -96,9 +96,6 @@ class ProcessingLevel:
 
 class BasicFileMetadata:
     """Metadata for basic files"""
-    file = None
-    site = ""
-
     def __init__(self, file, site):
         self.file = file
         self.site = site
@@ -127,16 +124,14 @@ class BasicFileMetadata:
 
 class I3FileMetadata(BasicFileMetadata):
     """Metadata for i3 files"""
-    processing_level = None
-    season_year = None
-    run = 0
-    subrun = 0
-    part = 0
-    meta_xml = dict()
-
     def __init__(self, file, site, processing_level):
         super().__init__(file, site)
         self.processing_level = processing_level
+        self.season_year = None
+        self.run = 0
+        self.subrun = 0
+        self.part = 0
+        self.meta_xml = dict()
         self._parse_filepath()
 
     def generate(self):
@@ -251,9 +246,6 @@ class I3FileMetadata(BasicFileMetadata):
 
 class L2FileMetadata(I3FileMetadata):
     """Metadata for L2 i3 files"""
-    gaps_dict = dict()
-    gcd_filepath = ""
-
     def __init__(self, file, site, dir_meta_xml, gaps_dict, gcd_filepath):
         super().__init__(file, site, ProcessingLevel.L2)
         self.meta_xml = dir_meta_xml
@@ -373,7 +365,6 @@ class L2FileMetadata(I3FileMetadata):
 
 class PFFiltFileMetadata(I3FileMetadata):
     """Metadata for PFFilt i3 files"""
-
     def __init__(self, file, site):
         super().__init__(file, site, ProcessingLevel.PFFilt)
         with tarfile.open(file.path) as tar:
@@ -411,18 +402,12 @@ class PFFiltFileMetadata(I3FileMetadata):
 
 class PFRawFileMetadata(I3FileMetadata):
     """Metadata for PFFilt i3 files"""
-
     def __init__(self, file, site):
         super().__init__(file, site, ProcessingLevel.PFRaw)
         with tarfile.open(file.path) as tar:
             for tar_obj in tar:
                 if ".meta.xml" in tar_obj.name:
                     self.meta_xml = xmltodict.parse(tar.extractfile(tar_obj))
-
-    def generate(self):
-        """Gather the file's metadata."""
-        metadata = super().generate()
-        return metadata
 
     def _parse_filepath(self):
         """Set the year, run, subrun, and part from the file name."""
@@ -453,13 +438,10 @@ class PFRawFileMetadata(I3FileMetadata):
 
 class MetadataManager:
     """Commander class for handling metadata for different file types"""
-    site = ""
-    dir_processing_level = None
-    l2_dir_metadata = dict()
-
     def __init__(self, path, site):
         self.site = site
         self.dir_processing_level = ProcessingLevel.from_path(path)
+        self.l2_dir_metadata = dict()
 
         if self.dir_processing_level == ProcessingLevel.L2:
             # get directory's metadata
@@ -568,13 +550,11 @@ def gather_file_info(dirs, site):
 async def request_post_patch(fc_rc, metadata):
     """POST metadata, and PATCH if file is already in the file catalog."""
     try:
-        logging.debug('POSTing...')
         _ = await fc_rc.request("POST", '/api/files', metadata)
         logging.debug('POSTed.')
     except requests.exceptions.HTTPError as e:
         if e.response.status_code == 409:
             patch_path = e.response.json()['file']  # /api/files/{uuid}
-            logging.debug('PATCHing...')
             _ = await fc_rc.request("PATCH", patch_path, metadata)
             logging.debug('PATCHed.')
         else:
--- a/resources/indexer.py
+++ b/resources/indexer.py
@@ -3,6 +3,7 @@
 
 import argparse
 import asyncio
+import collections
 import hashlib
 import logging
 import os
@@ -40,22 +41,24 @@ SEASONS = {
 }
 
 
-def _get_year_from_season(season):
-    if season:
-        for year, s in SEASONS.items():
-            if s == season:
-                return int(year)
-        raise Exception(f"No year found for season, {season}.")
+def _get_season_year(name):
+    """Return the year of the season start for the season's name."""
+    if name:
+        for season_year, n in SEASONS.items():
+            if n == name:
+                return int(season_year)
+        raise Exception(f"No season found for {name}.")
     else:
         return None
 
 
-def _get_season_from_year(year):
-    if year:
+def _get_season_name(season_year):
+    """Return the season's name for the year of the season start."""
+    if season_year:
         try:
-            return SEASONS[str(year)]
+            return SEASONS[str(season_year)]
         except KeyError:
-            raise Exception(f"No season name found for year, {year}.")
+            raise Exception(f"No season found for {season_year}.")
     else:
         return None
 
@@ -64,7 +67,7 @@ def _get_processing_level(file):
     """Return the processing level parsed from the file's dir path, case insensitive."""
     levels = {
         "PFRaw": "PFRaw",
-        "PFilt": "PFilt",
+        "PFFilt": "PFFilt",
         "PFDsT": "PFDsT",
         "Level2": "L2",
         "Level3": "L3",
@@ -74,7 +77,7 @@ def _get_processing_level(file):
     for level, level_for_metadata in levels.items():
         if level.upper() in dir_path_upper:
             return level_for_metadata
-    return ""
+    raise Exception(f"Cannot detect file's processing level, {file.path}.")
 
 
 def _get_run_number(file):
@@ -87,8 +90,10 @@ def _get_run_number(file):
     return run
 
 
-def _disect_filename(file):
-    """Return year/run/subrun/part number from filename"""
+def _disect_filepath(file):
+    """Return data_type, processing_level, year, run, subrun, & part number from file's path."""
+    # Ex: /data/exp/IceCube/2013/filtered/level2/1229/Run00123579_0/Level2_IC86.2013_data_Run00123579_Subrun00000004.i3.bz2
+    processing_level = _get_processing_level(file)
 
     def _get_year_from_year(file):
         # Ex: Level2_IC86.2017_[...]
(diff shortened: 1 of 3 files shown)


dsschult:
Looks like there are some conflicts with master.  Those need to get resolved.

resources/indexer.py
```diff
@@ -18,290 +19,292 @@
...
+            line = file.read(bufsize)
+            while line:
+                sha.update(line)
+                line = file.read(bufsize)
+        return sha.hexdigest()
+
+    def _get_locations(self):
+        """Return the locations object."""
+        if '.tar' in self.file.name:
+            return [{'site': self.site, 'path': self.file.path, 'archive': True}]
```
dsschult: This is not what the archive flag is for.  archive means things like the NERSC or DESY archive, which are not directly accessible.

resources/indexer.py
```diff
@@ -3,6 +3,7 @@
```
dsschult: Consider using a CVMFS metaproject with shebang so it auto-loads the env.
http://software.icecube.wisc.edu/documentation/info/cvmfs.html#self-contained-icetray-scripts

resources/indexer.py
```diff
@@ -339,53 +532,26 @@ def process_dir(path, site):
...
-                for tar_obj in tar:
-                    file = tar.extractfile(tar_obj)
-                    file_dict = yaml.load(file)
-                    # Ex. Level2_IC86.2017_data_Run00130484_Subrun00000000_00000188_gaps.txt
-                    name = tar_obj.name.split("_gaps.txt")[0]
-                    gaps_files[name] = file_dict
-        if "GCD" in dir_entry.name:  # Ex. Level2_IC86.2017_data_Run00130484_0101_71_375_GCD.i3.zst
-            run = _get_run_number(dir_entry)
-            gcd_files[str(run)] = dir_entry.path
+    manager = MetadataManager(path, site)
```
dsschult: Question: what happens when a directory isn't labelled correctly, but the files all start with PFRaw?  Does the manager assign the wrong level, and metadata is missed?

I'm wondering because of the loading that occurs off of disk every year, or other similar circumstances.

resources/indexer.py
ric-evans: What's the metaproject in this case?

resources/indexer.py
ric-evans: This is a requirement, the script would exit if the directory doesn't contain a processing level.

Is this problematic?

resources/indexer.py
dsschult: Either the latest release, or we maybe could use combo stable.

resources/indexer.py
dsschult: @jnbellinger how do you ingest files from disk?  Are you planning on using this script?

jnbellinger: I use "ltacmd catalog load" for each directory that I've identified as full.  I'm not sure if this gets called behind the scenes:  I don't see it in the LTA code.

dsschult: Checking LTA code, it looks like it does not call this.  Instead it does the bare minimum (just getting the checksum) and puts it directly in the file catalog.  I'm not sure if it's good or bad that those files don't have any other metadata.

jnbellinger: It is a bit generic.  For raw data, what extra benefit to we get from the metadata that isn't already encoded in the "PFRAw" and run number and whatever live tells us about the data quality? The beginning and end events; what else?

dsschult: I mean, in the LTA code you don't even know that it's a raw file, unless you look at the name.  This script at least adds fields for the processing level, run, subrun, file part, beginning and ending events, (maybe others?).

That's important if you're filtering for, say, all files that have to do with a run.  It's just not possible to look at the name of every file in the catalog and ask if some string is in the name (it would take forever, because there isn't a good way to index that query).

jnbellinger: I suppose one could do "after the fact" fixup for some of those things.  Or write something that simultaneously checksums and unpacks/reads the metadata file.  Doing a second read pass during the "catalog load" to pull the metadata and read it seems excessive.

Is the metadata available anywhere other than the bundled metadata file?

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+
+    def generate(self):
+        """Gather the file's metadata."""
+        metadata = dict()
+        metadata['logical_name'] = self.file.path
+        metadata['checksum'] = {'sha512': self.sha512sum()}
+        metadata['file_size'] = self.file.stat().st_size
+        metadata['locations'] = [{'site': self.site, 'path': self.file.path}]
+        metadata['create_date'] = date.fromtimestamp(os.path.getctime(self.file.path)).isoformat()
+        return metadata
```
blinkdog: I would rewrite this in a more literal style:
```
def generate(self):
    """Gather the file's metadata."""
    return {
        'logical_name': self.file.path,
        'checksum': {
            'sha512': self.sha512sum(),
        },
        'file_size': self.file.stat().st_size,
        'locations': [
            {
                'site': self.site,
                'path': self.file.path,
            },
        ],
        'create_date': date.fromtimestamp(os.path.getctime(self.file.path)).isoformat(),
    }
```

But I may also be scarred from years of JavaScript/CoffeeScript, so feel free to ignore.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+
+
+class I3FileMetadata(BasicFileMetadata):
+    """Metadata for i3 files"""
+    processing_level = None
+    season_year = None
+    run = 0
     subrun = 0
+    part = 0
+    meta_xml = dict()
```
blinkdog: I'd bring these lines into the constructor, preceding with `self.` as necessary.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+                'end_datetime': end_dt,
+                'first_event': first_event,
+                'last_event': last_event,
+                'event_count': event_count
+            }
+        return metadata
+
+    def _parse_filepath(self):
+        """Set the year, run, subrun, and part from the file name."""
+        pass
```
blinkdog: This function does not do what it says on the tin... is it used? Should it be removed?

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+    season_year = None
+    run = 0
     subrun = 0
+    part = 0
+    meta_xml = dict()
+
+    def __init__(self, file, site, processing_level):
+        super().__init__(file, site)
+        self.processing_level = processing_level
+        self._parse_filepath()
```
blinkdog: You are calling it here, but the implementation is empty.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+
+    @staticmethod
+    def parse_run_number(file):
+        """Return run number from filename."""
+        # Ex. Level2_IC86.2017_data_Run00130484_0101_71_375_GCD.i3.zst
+        # Ex: Level2_IC86.2017_data_Run00130567_Subrun00000000_00000280.i3.zst
+        # Ex: Run00125791_GapsTxt.tar
+        r = file.name.split('Run')[1]
+        run = int(r.split('_')[0])
+        return run
```
blinkdog: You've got a `file` attribute of your base class object (BasicFileMetadata), so it would seem natural to use it in your subclass (I3Metadata) instead of making this method static and passing it as an argument.

If you want to keep the static method, I'd refactor it out of the class and keep it as a stand-alone function.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+                        last = event_id
+        except:
+            status = "bad"
+        return first, last, count, status
+
+
+class L2FileMetadata(I3FileMetadata):
+    """Metadata for L2 i3 files"""
+    gaps_dict = dict()
+    gcd_filepath = ""
```
blinkdog: You set these in the constructor, so I'd just drop these lines.

resources/indexer.py
```diff
@@ -314,20 +312,209 @@ def _get_i3_part_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gc
...
+
+        else:
+            raise Exception(f"Filename not in a known L2 file format, {self.file.name}.")
+
+    @staticmethod
+    def is_file(file):
+        """True if the file is in the [...#].i3[...] file format."""
+        # Ex: Level2_IC86.2017_data_Run00130484_Subrun00000000_00000188.i3.zst
+        # check if last char of filename (w/o extension) is an int
+        return (".i3" in file.name) and (file.name.split('.i3')[0][-1]).isdigit()
```
blinkdog: You've got the `file` attribute from the grandparent superclass, so make this a non-static method perhaps?
Or refactor it out of the class into a stand-alone function?

resources/indexer.py
```diff
@@ -314,20 +312,209 @@ def _get_i3_part_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gc
...
+            self.part = int(p.split('.')[0])
+        else:
+            raise Exception(f"Filename not in a known PFFilt file format, {self.file.name}.")
+
+    @staticmethod
+    def is_file(file):
+        """ True if the file is in the [...#].tar.bz2 file format. """
+        # Ex. PFFilt_PhysicsFiltering_Run00131989_Subrun00000000_00000295.tar.bz2
+        # check if last char of filename (w/o extension) is an int
+        return (".tar.bz2" in file.name) and (file.name.split('.tar.bz2')[0][-1]).isdigit()
```
blinkdog: I'm pretty sure PFFilt files changed from `.tar.gz` to `.tar.bz2` at some point, so this may return the wrong thing if you're working with old/ancient PFFilt files.

resources/indexer.py
```diff
@@ -314,20 +312,209 @@ def _get_i3_part_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gc
...
+        super().__init__(file, site, ProcessingLevel.PFRaw)
+        with tarfile.open(file.path) as tar:
+            for tar_obj in tar:
+                if ".meta.xml" in tar_obj.name:
+                    self.meta_xml = xmltodict.parse(tar.extractfile(tar_obj))
+
+    def generate(self):
+        """Gather the file's metadata."""
+        metadata = super().generate()
+        return metadata
```
blinkdog: I don't think you need to override the `generate()` method here as it's just returning the result from the superclass?

resources/indexer.py
```diff
@@ -314,20 +312,209 @@ def _get_i3_part_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gc
...
+            self.part = int(p.split('.')[0])
+        else:
+            raise Exception(f"Filename not in a known PFRaw file format, {self.file.name}.")
+
+    @staticmethod
+    def is_file(file):
+        """ True if the file is in the [...#].tar.gz file format. """
+        # Ex. key_31445930_PFRaw_PhysicsFiltering_Run00128000_Subrun00000000_00000156.tar.gz
+        # check if last char of filename (w/o extension) is an int
+        return (".tar.gz" in file.name) and (file.name.split('.tar.gz')[0][-1]).isdigit()
```
blinkdog: I would refactor this out of the class to it's own function

resources/indexer.py
```diff
@@ -314,20 +312,209 @@ def _get_i3_part_file_metadata(file, site, run_meta_xml=None, gaps_dict=None, gc
...
+
+
+class MetadataManager:
+    """Commander class for handling metadata for different file types"""
+    site = ""
+    dir_processing_level = None
+    l2_dir_metadata = dict()
```
blinkdog: You set these in the constructor, so I would drop these lines.

resources/indexer.py
```diff
@@ -339,53 +526,26 @@ def process_dir(path, site):
...
-            run = _get_run_number(dir_entry)
-            gcd_files[str(run)] = dir_entry.path
+    manager = MetadataManager(path, site)
 
     # get files' metadata
     for dir_entry in scan:
         if dir_entry.is_symlink():
             continue
         elif dir_entry.is_dir():
+            logging.debug(f'Directory appended, {dir_entry.path}')
```
blinkdog: We want to keep this debug logging or remove it?

resources/indexer.py
```diff
@@ -339,53 +526,26 @@ def process_dir(path, site):
...
             dirs.append(dir_entry.path)
         elif dir_entry.is_file():
+            logging.debug(f'Gathering metadata for {dir_entry.name}...')
```
blinkdog: Again, keeping the debug logging or removing it?

resources/indexer.py
```diff
@@ -405,12 +565,29 @@ def gather_file_info(dirs, site):
...
+async def request_post_patch(fc_rc, metadata):
+    """POST metadata, and PATCH if file is already in the file catalog."""
+    try:
+        logging.debug('POSTing...')
+        _ = await fc_rc.request("POST", '/api/files', metadata)
+        logging.debug('POSTed.')
+    except requests.exceptions.HTTPError as e:
+        if e.response.status_code == 409:
+            patch_path = e.response.json()['file']  # /api/files/{uuid}
+            logging.debug('PATCHing...')
```
blinkdog: I would be specific in the log: PATCH {patch_path}

resources/indexer_make_dag.py
```diff
@@ -6,14 +6,22 @@
...
+    "PFDST": "unbiased/PFDST",
+    "PFRaw": "unbiased/PFRaw",
+}
 
-def _get_dirpaths():
-    years = [str(y) for y in range(2007, 2021)]
+
+def _get_dirpaths(begin_year, end_year, level):
+    years = [str(y) for y in range(begin_year, end_year)]
     dirs = [d for d in os.scandir(os.path.abspath('/mnt/lfs6/exp/IceCube')) if d.name in years]
```
blinkdog: Do you want to make this a configurable? Or leave it with the hard-coded path: `/mnt/lfs6/exp/IceCube`?

resources/indexer_make_dag.py
```diff
@@ -49,14 +60,14 @@ def main():
...
 """)
-        # request_memory = 2500
 
     dagpath = os.path.join(scratch, 'dag')
     with open(dagpath, 'w') as f:
-        for i, path in enumerate(_get_dirpaths()):
+        for i, path in enumerate(_get_dirpaths(args.begin, args.end, args.level)):
             f.write(f'JOB job{i} condor\n')
             f.write(f'VARS job{i} PATH="{path}"\n')
             f.write(f'VARS job{i} JOBNUM="{i}"\n')
```
blinkdog: Sorry, I don't know much about Condor or job batches, so I can't help much on if you've got the right format or not.

ric-evans: It's right; I've had no problems with what I've been running for L2 data.

resources/indexer.py
ric-evans: I've found it useful for debugging log files.

resources/indexer.py
ric-evans: I find it useful.

resources/indexer.py
ric-evans: I've removed this. It's already being logged by urllib3.connectionpool, so no need to double log.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
-def _get_processing_level(file):
-    """Return the processing level parsed from the file's dir path, case insensitive."""
-    levels = {
-        "PFRaw": "PFRaw",
-        "PFilt": "PFilt",
-        "PFDsT": "PFDsT",
-        "Level2": "L2",
-        "Level3": "L3",
-        "Level4": "L4"
+logging.basicConfig(level=logging.DEBUG)
```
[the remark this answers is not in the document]
dsschult: Good spotting, I agree with Patrick.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
 from icecube import dataclasses, dataio
 from rest_tools.client import RestClient
```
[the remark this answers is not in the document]
dsschult: Note that so far only Patrick is OCD about imports the way PEP8 wants.  I mostly just separate out python libraries from external libraries.  So I leave this up to you Ric.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+    @staticmethod
+    def name_to_year(name):
+        """Return the year of the season start for the season's name."""
+        if name:
+            for season_year, season_name in IceCubeSeason.SEASONS.items():
+                if season_name == name:
+                    return int(season_year)
+            raise Exception(f"No season found for {name}.")
+        else:
+            return None
```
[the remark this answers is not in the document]
dsschult: The other thing to think about with if statements, is that you want the easy branch to come first.

resources/indexer.py
ric-evans: Understandable, but I'm going to keep it. It helps readability-wise for me, maybe that's due to my Java upbringing.

resources/indexer.py
```diff
@@ -18,290 +19,287 @@
...
+    def sha512sum(self):
+        """Return the SHA512 checksum of the file given by path."""
+        bufsize = 4194304
+        sha = hashlib.new('sha512')
+        with open(self.file.path, 'rb', buffering=0) as file:
+            line = file.read(bufsize)
+            while line:
+                sha.update(line)
+                line = file.read(bufsize)
+        return sha.hexdigest()
```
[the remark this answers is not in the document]
dsschult: This looks ok to me.  In fact, it's very similar to code in IceProd:
https://github.com/WIPACrepo/iceprod/blob/682abe2fabdc03b69558bfabbbbdae36b91dee25/iceprod/core/functions.py#L123

resources/indexer.py
dsschult: I think the question comes down to if you ever want to optionally include a field, or if these fields are always present.  I'm fine with either style, though I will note that the proper way to create an empty dict is: `metadata = {}`

resources/indexer.py
dsschult: This is a good point.  Assume the compression can change, to include '.zst' or any other format.

resources/indexer.py
dsschult: I'm in favor of keeping debug logging if it is at all useful.  Then just run with a higher log level to filter it out in production.

resources/indexer.py
ric-evans: I made it a static method so it's more closely related to the class type. What would the benefit be to making this its own function?

resources/indexer.py
ric-evans: Good catch.

resources/indexer.py
dsschult: Note that it is idiomatic python to only put read-only variables at the class level, and declare all instance variables in `__init__()`.  This is mostly to make sure you're modifying the value you think you're modifying.

For instance, in this case you are modifying `self.l2_dir_metadata` in the `_prep_l2_dir_metadata` function.  But since you haven't declared it in the instance scope, it's actually modifying the class-level variable.  This could lead to interesting problems if there are multiple instances of this class at the same time.

ric-evans: Oh, I hadn't heard this before. I'll revise it.

resources/indexer.py
ric-evans: Done.

resources/indexer.py
ric-evans: I don't intend that an I3FileMetadata instance will be created, so `self._parse_filepath()` will always refer to a derived-class instance implementation. Is this not good practice?

resources/indexer.py
ric-evans: You metion this in https://github.com/WIPACrepo/file_catalog/pull/52#discussion_r376167193

resources/indexer.py
ric-evans: Good catch. In 2006 they're `.tar.gz`. I'll revise this.

resources/indexer.py
ric-evans: You mention this in https://github.com/WIPACrepo/file_catalog/pull/52#discussion_r376175831

resources/indexer.py
ric-evans: This is called prior to instantiation. I could move this logic to the constructor and throw an error, however that would be more verbose.

resources/indexer.py
ric-evans: Done.

resources/indexer.py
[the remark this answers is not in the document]
ric-evans: Good point.

resources/indexer.py
dsschult: Maybe it needs to raise `NotImplementedError`?  That implies that the base class method is purposely empty and should never be called.

resources/indexer.py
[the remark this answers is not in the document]
ric-evans: I'm using isort for this. I'd consider icecube and rest_tools imports third party in this case because they use independent source code.

resources/indexer.py
ric-evans: Is there a resolution here for the directory naming scheme?

I chose to look at the directory path name instead of the file name to handle older files like `/data/exp/IceCube/2008/filtered/level2/0324/PFFilt_PhysicsTrig_PhysicsFiltering_Run00110714_Level2_Part00000034.i3.gz`.

resources/indexer.py
ric-evans: @dsschult I like that.

resources/indexer.py
ric-evans: I like the compact style since there isn't a whole lot going on. I'll split the difference and replace `dict()` with `{}`.

resources/indexer.py
dsschult: @blinkdog any thoughts on this?

dsschult:
> There is a lot of functionality here that deals with lots of fiddly parsing details.
> I'd recommend writing some unit tests to help flush out any bugs.

I'm unsure how much unit tests will help, especially since you need the actual files to get useful metadata.  But if we do go that route, this stops becoming a helper script and becomes part of the main project (or a separate project).

dsschult:
At this point most of my complaints are addressed, and I'll list it as approved.

resources/indexer.py
ric-evans: I'm adding support for `.tar.gz`, `.tar.bz2`, `.tar.zst`.

resources/indexer.py
ric-evans: I resolved this by raising `NotImplementedError`.

resources/indexer.py
blinkdog: On the issue of metadata in the catalog, `ltacmd catalog load` should probably be deprecated in favor of this script after it is merged. I needed a quick way for the operator to get entries in the catalog for production deployment testing, but really there's no substitute for doing it right the the first time. This script is 'doing it right' in terms of catalog metadata.

resources/indexer.py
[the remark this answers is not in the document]
ric-evans: That's funny, usually I write in null-guardrails (None-guardrails?) like you suggested. I'll blame that on copy-and-pasting when I was refactoring.

resources/indexer.py
blinkdog: On the issue of file identification, I would look simply for `PFRaw`, `PFFilt`, and `PFDST` anywhere within the full path. There should be no data files anywhere that have multiple tags like this in the path. A case like that would mean the file is in the wrong directory, or the file has a _very_ strange name (PFRaw_PFFilt_blah_blah), and both are something we'd probably like an operator to correct before we start filling the catalog with metadata about them.

As for L2, L3, L4, I'm not really familiar with the unique substring tag that would identify a file of that type. I imagine there's got to be one for each, but I haven't worked with those files very often, so I'm in the dark.

resources/indexer_make_dag.py
blinkdog: @ric-evans Did you want to make this root directory something you can specify from the command-line?

blinkdog:
The only comment I'd really like to see a decision on is: https://github.com/WIPACrepo/file_catalog/pull/52#discussion_r376178609

The rest looks very good, and ready to merge.

resources/indexer_make_dag.py
ric-evans: Sure, I'll ad that as an option.

resources/indexer.py
dsschult: I think we should leave as-is for this pull request, but think of this for a new pull request.

jnbellinger: I've not seen a unique substring tag for L2/L3/L4 either.  
And I agree that parsing the directory and file name is probably good enough.  Every now and then we got duplicate run numbers.  In the past this hasn't been a problem, since one or more of the duplicates were test runs.  I assume that will continue to be the reason for such glitches in the future, and thus not a serious problem.
Will we want start/stop times first/last  event numbers in the catalog too?  Once upon a time I thought that might be good to have for coincidence studies with other detectors, but after seeing the time windows they use, I suspect just having the right run number might be good enough for that work.

PR #56, opened 2020-03-11
increase indexer performance for indexing all of /data/exp/

ric-evans:
Overview:

**indexer.py**
- user can specify paths of directories and files (previously just directories)
- user can also specify paths via a file where each line is a path (previously all paths had to be command-line arguments)
- REST POSTing is done in real-time (previously this was done after all metadata collecting finished)
- smarter multiprocessing: user can specify the number of processes to utilize. the file indexing is internally load-balanced, ie. each process will make nearly the same number of REST POSTs each
- user can define the logging level
- user can specify paths to skip via a file containing paths (--blacklist-file)

**directory_scanner.py**
- new script for getting a list of all files in a path
- file paths are outputted to stdout
- the number of files and any errors are logged to stderr

**indexer_make_dag.py**
- added 'Everything' level to index all of /data/exp/
    + directory_scanner.py is used
    + many subprocess.check_call()'s get that list of paths to sorted order and split into files with 10000 lines/paths each, each file to be processed by a condor job
    + this staging work is done in /data/user/{user}/indexer-data-exp/
    + the whole list of paths is placed in a file for future reference, see below
    + using the --previous-data-exp option (with a previously-generated file of paths) will only index any new files
- added dry-run argument for testing


resources/indexer/directory_scanner.py
```diff
@@ -0,0 +1,63 @@
...
+import logging
+import os
+from concurrent.futures import ProcessPoolExecutor
+from time import sleep
+
+
+def process_dir(path):
+    """Print out file paths and return sub-directories."""
+    try:
+        scan = list(os.scandir(path))
```
dsschult: Is there a reason for this to be a list?  It looks like the default return value of a generator would work also.

resources/indexer/directory_scanner.py
```diff
@@ -0,0 +1,63 @@
...
+    except (PermissionError, FileNotFoundError):
+        scan = []
+    dirs = []
+
+    all_file_count = 0
+    for dir_entry in scan:
+        if dir_entry.is_symlink():
```
dsschult: Should also check for other types of files, like pipes or devices. Those should be skipped as well.

resources/indexer/directory_scanner.py
```diff
@@ -0,0 +1,63 @@
...
+        while futures or dirs:
+            for d in dirs:
+                futures.append(pool.submit(process_dir, d))
+            while not futures[0].done():
+                sleep(0.1)
+            future = futures.pop(0)
+            dirs, result_all_file_count = future.result()
+            all_file_count = all_file_count + result_all_file_count
+
+    logging.info(f"\nFile Count: {all_file_count}")
```
dsschult: Not sure the newline does what you want, or is necessary.

resources/indexer/indexer.py
```diff
@@ -670,14 +635,140 @@ def gather_file_info(dirs, site, basic_only=False):
...
+
+
+def sorted_unique(infile, others=None):
+    """Read in lines from file, aggregate with those in others list. Return all unique lines, sorted."""
+    lines = []
+    if others:
+        lines.extend(others)
+    if infile:
+        with open(infile) as file:
+            lines.extend([l.rstrip() for l in file])
```
dsschult:
```suggestion
            lines.extend(l.rstrip() for l in file)
```
You shouldn't need to make the list, and can pass the generator to `extend`.

resources/indexer/directory_scanner.py
```diff
@@ -0,0 +1,63 @@
...
+            continue
+        elif dir_entry.is_dir():
+            dirs.append(dir_entry.path)
+        elif dir_entry.is_file():
+            all_file_count = all_file_count + 1
+            if not dir_entry.path.strip():
+                logging.info(f"Blank file name in: {os.path.dirname(dir_entry.path)}")
+            else:
+                try:
+                    print(dir_entry.path)
```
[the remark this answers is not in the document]
dsschult: The "output" of the program is a series of paths on stdout, so this should be correct.

resources/indexer/indexer.py
```diff
@@ -670,14 +635,140 @@ def gather_file_info(dirs, site, basic_only=False):
...
     parser = argparse.ArgumentParser(description='Find files under PATH(s), compute their metadata and '
                                      'upload it to File Catalog.',
                                      epilog='Notes: (1) symbolic links are never followed.',
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    parser.add_argument('path', metavar='PATH', nargs='+',
+    parser.add_argument('paths', metavar='PATHS', nargs='*',
                         help='path(s) to scan for files.')
+    parser.add_argument('--paths-file', dest='paths_file', default=None,
+                        help='file containing path(s) to scan for files. (use this option for a large number of paths)')
+    parser.add_argument('--processes', type=int, default=1,
```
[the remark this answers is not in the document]
dsschult: Since it's processing in condor, a default of 1 is actually fairly good.  You could technically look in the condor classad for the number of cpus allocated.

Asking for the number of cpus on the machine is probably the worst approach when running in condor.

resources/indexer/directory_scanner.py
ric-evans: Good point.

resources/indexer/directory_scanner.py
ric-evans: That was left-over from some refactoring, I'll remove it.

resources/indexer/indexer_make_dag.py
```diff
@@ -40,46 +92,92 @@ def main():
...
     args = parser.parse_args()
 
-    scratch = f"/scratch/eevans/{args.level}indexer"
+    # check paths in args
+    for path in [args.env, args.blacklist, args.previous_data_exp]:
+        if path and not os.path.exists(path):
+            raise FileNotFoundError(path)
+
+    # make condor scratch directory
+    scratch = os.path.join('/scratch/', getpass.getuser(), f'{args.level}indexer')
```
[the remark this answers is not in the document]
ric-evans: Sure thing

resources/indexer/indexer.py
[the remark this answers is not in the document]
ric-evans: As @dsschult touched on, this is meant to be defined by the parent/condor script that runs it

resources/indexer/indexer.py
```diff
@@ -670,14 +635,140 @@ def gather_file_info(dirs, site, basic_only=False):
...
+                continue
+
+    return sub_files
+
+
+def path_in_blacklist(path, blacklist):
+    """Return True if path is in the blacklist."""
+    for b in blacklist:
+        if path.startswith(b):
+            logging.debug(f'Skipping {path}, file and/or directory is in blacklist.')
```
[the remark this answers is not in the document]
ric-evans: That's a good suggestion

resources/indexer/directory_scanner.py
ric-evans: I'm looking at https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_dir. Do you know if there's a type hierarchy here, ie. `(path.is_file() and path.is_fifo()) is True`?

dsschult: Most of them are probably also files, but otherwise there isn't much of a hierarchy.

blinkdog: My read of the documentation is that each of those are orthogonal, with the exception that a symlink pointing to a regular file will be true for both `is_file` and `is_symlink`. Otherwise it should be a choose-one between them.

If this is how it actually works in practice... :man_shrugging: you may need to write some unit tests, depending on how important it is.

dsschult: The key ones are socket and fifo, so just adding them to the if statement should be fine.

resources/indexer/indexer.py
```diff
@@ -670,14 +635,140 @@ def gather_file_info(dirs, site, basic_only=False):
...
+    logging.debug(f'{filepath} gathered.')
+    logging.debug(metadata)
+    await request_post_patch(fc_rc, metadata, no_patch)
+
+
+async def process_paths(paths, manager, fc_rc, no_patch):
+    """POST metadata of files given by paths, and return any directories."""
+    sub_files = []
+
+    paths = [p for p in paths if not os.path.islink(p)]  # remove symbolic links
```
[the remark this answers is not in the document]
ric-evans: check out `is_processable_path()`

ric-evans:
> It mostly looks good. Just some small suggestions.
> 
> As for the future, this is starting to get large enough we may want to place it into the actual file catalog code, maybe as a client.

Noted.

dsschult:
looks good to me

PR #57, opened 2020-04-08
Offload directory scanning

ric-evans:
- moved the /data/exp filepath collection process from indexer_make_dag.py to all_paths_.py and all_paths_make_condor.py

- now in order to index all of /data/exp:
1. run all_paths_make_condor.py
2. run indexer_make_dag.py with flag pointing to directory produced from all_paths_.py


resources/indexer/all_paths.py
```diff
@@ -0,0 +1,77 @@
...
+        check_call_print(f'rm {file_orig}'.split())  # Cleanup
+
+        # Get lines(file paths) unique to this scan versus the previous file
+        if previous:
+            check_call_print(f'comm -1 -3 {previous} {file_sort} > {file_sort}.unique', shell=True)
+            check_call_print(f'mv {file_sort}.unique {file_sort}'.split())
+
+        # split the file into n files
+        check_call_print(f'mkdir {dir_split}'.split())
+        check_call_print(f'split -l{paths_per_file} {file_sort} paths_file_'.split(), cwd=dir_split)
```
dsschult, line 40: This entire block of subprocess calls leaves a bad taste in my mouth, (it's very non-pythonic).  I'll let it slide because it works, but a pure-python solution for these operations isn't actually that hard.

dsschult:
I'm ok with "squash and merge" now.

resources/indexer/all_paths.py
ric-evans: Ahh I can see that

PR #58, opened 2020-04-13
add file count ability

dsschult:


file_catalog/server.py
```diff
@@ -408,7 +408,10 @@ def get(self):
             logging.warn('query parameter error', exc_info=True)
             self.send_error(400, message='invalid query parameters')
             return
-        files = yield self.db.find_files(**kwargs)
+        if 'count' in kwargs:
+            files = yield self.db.count_files(**kwargs)
```
ric-evans, line 412: I don't think that dynamically changing the type of `files` is intuitive. I expected `files` to always be a list.

tests/test_files.py
```diff
@@ -44,6 +44,30 @@ def test_10_files(self):
...
+            u'locations': [{u'site':u'test',u'path':u'blah.dat'}]
+        }
+        data = r.request_seq('POST', '/api/files', metadata)
+        self.assertIn('_links', data)
+        self.assertIn('self', data['_links'])
+        self.assertIn('file', data)
+        url = data['file']
+        uid = url.split('/')[-1]
+
+        data = r.request_seq('GET', '/api/files?count=1')
```
blinkdog, line 65: Per discussion in #dev-sprint, let's make this `/api/files/count?query=...`

dsschult: And done.

blinkdog:
I like the refactoring of the query building.

PR #59, opened 2020-04-20
Mongo uri

ric-evans:


file_catalog/mongo.py
```diff
@@ -2,25 +2,35 @@
...
+    def __init__(self, host=None, port=None, authSource=None, username=None, password=None, uri=None):
+
+        if uri:
+            logger.info(f"MongoClient args: uri={uri}")
+            self.client = MongoClient(uri, authSource=authSource).file_catalog
+        else:
+            logger.info('MongoClient args: host=%s, port=%s, username=%s', host, port, username)
+            self.client = MongoClient(host=host, port=port,
+                                      authSource=authSource,
+                                      username=username, password=password).file_catalog
```
dsschult, line 33: I wonder what happens if you don't pass one of these variables in the environment.  What kind of error message does it give?  Probably not very informative for fixing the problem.

file_catalog/mongo.py
```diff
@@ -2,25 +2,35 @@
...
 
         self.client = MongoClient(host=host, port=port,
                                   authSource=authSource,
```
blinkdog, line 36: Does the line override what was set in the if/else block above?

dsschult: It looks like it does.  That seems like a copy/paste error?

file_catalog/mongo.py
ric-evans: I assume, the Mongo Client would fail to connect. How can I test this?

dsschult: I guess write a test missing an arg, that is expected to fail?

dsschult: I don't think this is a blocker, just a note that it produces poor error messages in this case.

PR #66, opened 2021-02-03
Updates to Indexes

ric-evans:
It may also be useful to add indexes for `simulation`'s values. However, I don't know which.


dsschult:
I'm not exactly sure either.  Dataset is the only field that sticks out.  I bet we find others whenever we start letting users access the service.

PR #67, opened 2021-02-04
Logging Format Updates

ric-evans:
Biggest bonus is that coloredlogs includes a timestamp which will make debugging easier

### DIFF
--- a/file_catalog/__main__.py
+++ b/file_catalog/__main__.py
@@ -4,6 +4,7 @@ import argparse
 import logging
 import os
 
+import coloredlogs  # type: ignore[import]
 from file_catalog.server import Server
 from file_catalog.config import Config
 from pprint import pprint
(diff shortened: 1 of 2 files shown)


dsschult:
My feelings about colored logs are "meh." (I can change the log formatting to include a timestamp in the python logger too)

But I won't object to it if you want to use colored logs.

ric-evans:
> My feelings about colored logs are "meh." (I can change the log formatting to include a timestamp in the python logger too)
> 
> But I won't object to it if you want to use colored logs.

k8s logs won't forward the coloring unfortunately (yet?). But, the actually formatting is nice, even though that can be done with just `logging.basicConfig()`

blinkdog:
Ahhh coloring, takes me back to the days of crafting [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) to make output pretty in dial-up modem terminal programs and MS-DOS.

This does make logs harder to use in plain text editors, but [there are ways to filter ANSI codes](https://stackpointer.io/unix/unix-linux-remove-ansi-escape-sequences/464/) so heck, why not? :smile:

blinkdog (approved):

dsschult:
note that `coloredlogs` is aware of the terminal type and won't input ASNI codes into non-interactive terminals (unless you really, really tell it to)

PR #74, opened 2021-02-11
Add 'all-keys' REST Query Parameter

ric-evans:
I've added an `"all-keys"` REST query parameter which essentially overrides `"keys"` by returning the FC files with their entire metadata. This new parameter is intended to be a bool/flag. All enhancements are backwards compatible and tested.

### This has several advantages:
1. Performance is increased **considerably** when querying `/api/files` versus multiple queries to `api/file/<uuid>`
2. This doesn't require the REST client/user to know what fields are available. That's useful since the file metadata schema varies among file types.
3. Individual REST routes can maintain their default keys when no `"keys"` parameter is included. Eg: `api/file/<uuid>` provides `"logical_name"` and `"uuid"` by default.

### Notes:
- `argbuilder.py` includes utility functions for constructing common arguments that are first parsed by `urlargparse.parse()`
-  `"all-keys"` is available for all routes where `"keys"` is. See `argbuilder.py`.
     + _only `/api/files/` has been tested so far_
- `mongo.py` has been updated, as well as type-hinted, formatted, and documented
     + an `AllKeys` instance is used as an indicator and to prevent convoluted multiple parameter logic.

### DIFF
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -81,7 +81,7 @@ class Mongo(object):
             if default:  # use default keys if they're available
                 projection.update(default)
         elif isinstance(keys, AllKeys):
-            pass  # only use "id_" constraint in projection
+            pass  # only use "_id" constraint in projection
         elif isinstance(keys, list):
             projection.update({k: True for k in keys})
         else:
@@ -121,7 +121,7 @@ class Mongo(object):
     ) -> List[Dict[str, Any]]:
         """Find files.
 
-        Optionally, apply keyword arguments. "id_" is always excluded.
+        Optionally, apply keyword arguments. "_id" is always excluded.
 
         Decorators:
             run_on_executor
@@ -224,7 +224,7 @@ class Mongo(object):
     ) -> List[Dict[str, Any]]:
         """Find all collections.
 
-        Optionally, apply keyword arguments. "id_" is always excluded.
+        Optionally, apply keyword arguments. "_id" is always excluded.
 
         Decorators:
             run_on_executor
@@ -271,7 +271,7 @@ class Mongo(object):
     ) -> List[Dict[str, Any]]:
         """Find snapshots.
 
-        Optionally, apply keyword arguments. "id_" is always excluded.
+        Optionally, apply keyword arguments. "_id" is always excluded.
 
         Decorators:
             run_on_executor
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -77,10 +77,11 @@ class Mongo(object):
         default: Optional[Dict[str, bool]] = None,
     ) -> Dict[str, bool]:
         projection = {"_id": False}
-        if not keys and default:
-            projection.update(default)
+        if not keys:
+            if default:  # use default keys if they're available
+                projection.update(default)
         elif isinstance(keys, AllKeys):
-            pass
+            pass  # only use "id_" constraint in projection
         elif isinstance(keys, list):
             projection.update({k: True for k in keys})
         else:
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -1,26 +1,28 @@
+"""File Catalog MongoDB Interface."""
+
+
 from __future__ import absolute_import, division, print_function
 
 import datetime
 import logging
 from concurrent.futures import ThreadPoolExecutor
+from typing import Any, cast, Dict, List, Optional, Union
 
-import pymongo
-from bson.objectid import ObjectId
+import pymongo  # type: ignore[import]
 from pymongo import MongoClient
-from pymongo.errors import BulkWriteError
 from tornado.concurrent import run_on_executor
 
-try:
-    from collections.abc import Iterable
-except ImportError:
-    from collections import Iterable
+logger = logging.getLogger("mongo")
 
 
+class AllKeys:  # pylint: disable=R0903
+    """Include all keys in MongoDB find*() methods."""
 
-logger = logging.getLogger('mongo')
 
 class Mongo(object):
-    """A ThreadPoolExecutor-based MongoDB client"""
+    """A ThreadPoolExecutor-based MongoDB client."""
+
+    # fmt:off
     def __init__(self, host=None, port=None, authSource=None, username=None, password=None, uri=None):
 
         if uri:
@@ -40,12 +42,12 @@ class Mongo(object):
         self.client.files.create_index([('locations.site',pymongo.DESCENDING),('locations.path',pymongo.DESCENDING)], background=True)
         self.client.files.create_index('locations.archive', background=True)
         self.client.files.create_index('create_date', background=True)
-        
+
         # all .i3 files
         self.client.files.create_index('content_status', sparse=True, background=True)
         self.client.files.create_index('processing_level', sparse=True, background=True)
         self.client.files.create_index('data_type', sparse=True, background=True)
-        
+
         # data_type=real files
         self.client.files.create_index('run_number', sparse=True, background=True)
         self.client.files.create_index('start_datetime', sparse=True, background=True)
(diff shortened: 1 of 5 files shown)


file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
+            logger.warning(
+                "Cannot determine if document has been modified since `result.modified_count` has the value `None`. `result.matched_count` is %s",
+                result.matched_count,
+            )
         elif result.modified_count != 1:
-            logger.warn('updated %s files with id %r',
-                        result.modified_count, uuid)
-            raise Exception('did not update')
+            logger.warning("updated %s files with id %r", result.modified_count, uuid)
+            raise Exception("did not update")
```
blinkdog, line 186: `"did not update"` doesn't match so well with the logging statement above it.
Consider building a string and using it in both the logger warning and the exception message?

file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
```
blinkdog, line 205: `"did not update"` doesn't match so well with the logging statement above it.
Consider building a string and using it in both the logger warning and the exception message?

file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
         result = self.client.files.delete_one(filters)
 
         if result.deleted_count != 1:
-            logger.warn('deleted %d files with filter %r',
-                        result.deleted_count, filter)
-            raise Exception('did not delete')
+            logger.warning(
+                "deleted %d files with filter %r", result.deleted_count, filters
+            )
+            raise Exception("did not delete")
```
blinkdog, line 216: `"did not delete"` doesn't match so well with the logging statement above it.
Consider building a string and using it in both the logger warning and the exception message?

file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
-
+    def find_collections(
+        self,
+        keys: Optional[Union[List[str], AllKeys]] = None,
+        limit: Optional[int] = None,
+        start: int = 0,
+    ) -> List[Dict[str, Any]]:
+        """Find all collections.
+
+        Optionally, apply keyword arguments. "id_" is always excluded.
```
blinkdog, line 227: Should be `"_id" is always excluded.` ?

file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
+    def find_snapshots(
+        self,
+        query: Optional[Dict[str, Any]] = None,
+        keys: Optional[Union[List[str], AllKeys]] = None,
+        limit: Optional[int] = None,
+        start: int = 0,
+    ) -> List[Dict[str, Any]]:
+        """Find snapshots.
+
+        Optionally, apply keyword arguments. "id_" is always excluded.
```
blinkdog, line 274: Should be `"_id" is always excluded.` ?

file_catalog/mongo.py
```diff
@@ -229,12 +321,14 @@ def append_distinct_elements_to_file(self, uuid, metadata):
...
+            logger.warning(
+                "Cannot determine if document has been modified since `result.modified_count` has the value `None`. `result.matched_count` is %s",
+                result.matched_count,
+            )
         elif result.modified_count != 1:
-            logger.warn('updated %s files with id %r',
-                        result.modified_count, uuid)
-            raise Exception('did not update')
+            logger.warning("updated %s files with id %r", result.modified_count, uuid)
+            raise Exception("did not update")
```
blinkdog, line 334: `"did not update"` doesn't match so well with the logging statement above it.
Consider building a string and using it in both the logger warning and the exception message?

tests/test_files.py
```diff
@@ -68,6 +73,76 @@ def test_11_files_count(self):
...
+        url = data["file"]
+        uid = url.split("/")[-1]
+
+        # w/o all-keys
+        data = r.request_seq("GET", "/api/files")
+        assert set(data["files"][0].keys()) == {"logical_name", "uuid"}
+
+        # w/ all-keys
+        body = {"all-keys": True}
+        data = r.request_seq("GET", "/api/files", body)
```
blinkdog, line 105: GET Request + Request Body are like ketchup and vanilla ice-cream, individually great but don't belong together.
https://stackoverflow.com/a/983458

I know this is just a test case, and I actually like the clean look; it's easy to tell which query parameters you're setting (and what their values are) without mentally parsing the query string for a question mark.

I'd be happy with a comment that said "`# body shouldn't be sent to a GET request, but we're only testing so we'll give ourselves a pass`" or you can rewrite the query strings.

blinkdog (approved):
LGTM. You've consolidated and cleaned up a lot of code here.
I appreciate the test cases showing the interactions between `keys` and `all-keys` in their various forms.
My review has some suggestions, but nothing that should prevent a merge.

file_catalog/mongo.py
```diff
@@ -93,134 +112,207 @@ def find_files(self, query={}, keys=None, limit=None, start=0):
...
+    def find_files(
+        self,
+        query: Optional[Dict[str, Any]] = None,
+        keys: Optional[Union[List[str], AllKeys]] = None,
+        limit: Optional[int] = None,
+        start: int = 0,
+    ) -> List[Dict[str, Any]]:
+        """Find files.
+
+        Optionally, apply keyword arguments. "id_" is always excluded.
```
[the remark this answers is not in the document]
ric-evans: yup

file_catalog/mongo.py
ric-evans: sure thing

file_catalog/mongo.py
ric-evans: sure thing

file_catalog/mongo.py
ric-evans: sure thing

file_catalog/mongo.py
ric-evans: you got it

tests/test_files.py
ric-evans: While it looks like [Roy agrees with you](https://stackoverflow.com/a/983458/13156561), our ice cream has been ketchup-y since the beginning: https://github.com/WIPACrepo/file_catalog/blame/64dc55510308a0fa98145b3d1c2d7628bbbb4fcf/file_catalog/server.py#L384

blinkdog: Yeah, I don't think you need to rewrite `request_seq` or anything deep here.

I'm just suggesting to add a comment to the test to show that it's not recommended practice; it's just a test, not production code, so a little cheating is fine in this context.

Or, rewrite the test to put the parameters into the query string and remove body altogether. This second option requires more work and (IMHO) obscures the purpose of the test, so I think a comment is better.

Trying to fix it so no GET requests can possibly have a request body is way beyond the scope of this PR; please don't do that here. :smile:

dsschult: I think there's some confusion here about how the rest client works.  The third function parameter is for arguments, and is not directly a message body.  If it is a GET or HEAD request, they get converted to query parameters.  Otherwise, they are sent as json in the body.

So, I think the only thing needed here is renaming the variable from `body` to something like `args`.

blinkdog: Ahhh... that works too!

ric-evans: Thanks for the clarification @dsschult. I've always thought about that argument as the "body" of the _rest_tools_ request, and _not_ the `body` of the REST request. At any rate, I renamed the variable.

Issue #68, opened 2021-02-08
Cleanup File Catalog Near-Duplicates


dsschult:
There are entries in the file catalog with filepaths rooted at `/mnt/lfs*/` which is not good for longevity

ric-evans:
closed by https://github.com/WIPACrepo/file-catalog-indexer/pull/12

jnbellinger:
What do you mean by the "x2"?  Do you mean all 1679  /mnt/lfs7 entries?  The /data/exp entries should be fine

ric-evans:
@jnbellinger responded to you over here: https://github.com/WIPACrepo/file_catalog/issues/98#issuecomment-810641751

Issue #73, opened 2021-02-11
Cast offline_processing_metadata.season Values to Integers


ric-evans:
The FC Indexer v1 used strings, v2 uses ints (eg: "2021" vs 2021). This needs to be cleaned up.

PR #75, opened 2021-02-16
Cleanup Scripts

ric-evans:
closes https://github.com/WIPACrepo/file_catalog/issues/68 and https://github.com/WIPACrepo/file_catalog/issues/73

##### `deduplicate.py`
- script for finding & deleting duplicate File Catalog records (aka evil twins), where one is listed as the absolute path (Ex: rooted at /data/exp/) and the other is the realpath (rooted at /mnt/lfs6/exp/)

##### `cast_season_years.py` 
- script for finding & fixing File Catalog records where the `"offline_processing_metadata.season"` (year) is a string. The "offline_processing_metadata" object is patched with an int-cast season value.


ric-evans:
relies on https://github.com/WIPACrepo/file_catalog/pull/74

ric-evans:
relies on https://github.com/WIPACrepo/file_catalog/pull/76

ric-evans:
`deduplicate.py` is done

resources/cleanup_scripts/deduplicate.py
```diff
@@ -0,0 +1,300 @@
...
+    previous_page: List[Dict[str, Any]] = []
+    page = 0
+    while True:
+        logging.info(
+            f"Looking for more bad-rooted paths (page={page}, limit={PAGE_SIZE})..."
+        )
+
+        # Query
+        body = {"start": page * PAGE_SIZE, "limit": PAGE_SIZE, "all-keys": True}
+        resp = rc.request_seq("GET", "/api/files", body)
```
dsschult, line 210: Would it be faster to do the logical name query here, so it happens on the server side?  There should be a way to send a mongo query to the server to ask for only prefixed logical names.

dsschult (approved):
Generally looks good

resources/cleanup_scripts/deduplicate.py
ric-evans: Is that possible?

ric-evans:
`cast_season_years.py` is ready for review

resources/cleanup_scripts/deduplicate.py
dsschult:
```python
rc.request_seq('GET', '/api/files', {'limit': 1, 'query':json.dumps({"logical_name":{ '$regex': '^\/mnt\/lfs.*' }})})
```

ric-evans: I didn't know mongo supported regex--that cleans things up nicely. 

Now I'm thinking that adding a "directory" base argument (which would essentially make a `$regex` query on the path name) would be useful.

dsschult: Sounds like another PR :smile:

ric-evans: https://github.com/WIPACrepo/file_catalog/issues/77

PR #80, opened 2021-03-01
Metadata Typing

ric-evans:
- added `schema/types.py` containing metadata type hints
     + this can be used by clients (aka the indexer)
     + this could allow `validation.py`  to validate metadata using introspection (future)
- moved `validation.py` to `schema/validation.py`


file_catalog/schema/types.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # i3 File:
+    data_type: Optional[str]
+    processing_level: Optional[str]
+    content_status: str
+
+    # /data/exp/* i3 File:
+    software: Optional[List[SoftwareEntry]]
+    run: Run
+    offline_processing_metadata: OfflineProcessingMetadata
```
dsschult, line 178: Are we really using such a long name?

dsschult (commented):
Type information generally looks good.  I guess one question is how to fit the rest of the columns in the metadata doc.

file_catalog/schema/types.py
ric-evans: Yup, I think this is the longest name

ric-evans:
> Type information generally looks good. I guess one question is how to fit the rest of the columns in the metadata doc.

We can put descriptions as an inline comment. For whether the key is required, that would need to be an independent dict.

ric-evans:
> > Type information generally looks good. I guess one question is how to fit the rest of the columns in the metadata doc.
> 
> We can put each description as an inline comment. For whether the key is required, that would need to be an independent dict.

We could really go all out and duplicate/migrate the Google Doc into a dict/yaml. But this is beyond the scope of this PR

dsschult:
> > > Type information generally looks good. I guess one question is how to fit the rest of the columns in the metadata doc.
> > 
> > 
> > We can put each description as an inline comment. For whether the key is required, that would need to be an independent dict.
> 
> We could really go all out and duplicate/migrate the Google Doc into a dict/yaml. But this is beyond the scope of this PR

Fair enough.

ric-evans:
> > > > Type information generally looks good. I guess one question is how to fit the rest of the columns in the metadata doc.
> > > 
> > > 
> > > We can put each description as an inline comment. For whether the key is required, that would need to be an independent dict.
> > 
> > 
> > We could really go all out and duplicate/migrate the Google Doc into a dict/yaml. But this is beyond the scope of this PR
> 
> Fair enough.

https://github.com/WIPACrepo/file_catalog/issues/81

PR #85, opened 2021-03-01
switch docker image base to python:3.8

dsschult:


Dockerfile
```diff
@@ -1,13 +1,11 @@
-FROM alpine:3.10
-
-RUN apk add --no-cache gcc git libffi-dev musl-dev openssl-dev python3-dev
-RUN pip3 install --upgrade pip # as of alpine:3.10, /usr/bin/pip won't exist until this command runs
+FROM python:3.8
```
ric-evans, line 1: do you want to go with `python:latest`?

dsschult: I prefer pinning the major:minor version, and only updating as needed.  That way suddenly a test doesn't fail because we went up a minor version and something in python changed.

Issue #77, opened 2021-02-26
Enable Querying by Directory for /api/files


ric-evans:
Adding a `"directory"` base argument (which would essentially make a `$regex` query on the path-name) would be useful.

This could support the `"logical_name"` path an/or `"locations"` paths.

ric-evans:
This will return files (because `/api/files`), but a similar solution could be used to get sub-directories (`/api/dirs`?). I don't know how useful that would be.

dsschult:
> This will return files (because `/api/files`), but a similar solution could be used to get sub-directories (`/api/dirs`?). I don't know how useful that would be.

My opinion is we shouldn't try to look like a filesystem, because we are not.  Users should mostly rely on metadata queries.

I view the "directory" argument as a helper for our use in retrieving multiple files, not a way to navigate a posix filesystem.

PR #87, opened 2021-03-03
Add Path-Like Arguments for Querying Files

ric-evans:
Introduces support for shortcut/base request arguments:
- `"directory"` and/or `"filename"`
     - query by directory path and/or the filename
     - Ex: `{"directory": "/foo/bar"}` -> `["/foo/bar/ham.i3", "/foo/bar/green.i3", "/foo/bar/eggs/ham.i3"]`
     - Ex: `{"directory": "/foo/bar", "filename": "ham.i3"}` -> `["/foo/bar/ham.i3", "/foo/bar/eggs/ham.i3"]`
     - Ex: `{"filename": "ham.i3"}` -> `["/foo/bar/ham.i3", "/foo/bar/eggs/ham.i3", "/beef/pork/ham.i3"]`
- `"path"` 
     - query by filepath (currently just a pseudonym for `"logical_name"`)
- `"path-regex"` 
     - query by regex

#### Order of precedence (decreasing): 
1.  `"directory"` & `"filename"`
1.  `"logical_name"` 
2. `"path"`
5. `"path-regex"` 
6. `"logical_name"` sub-field within the `"query"` dict

closes https://github.com/WIPACrepo/file_catalog/issues/77

### DIFF
--- a/file_catalog/argbuilder.py
+++ b/file_catalog/argbuilder.py
@@ -30,7 +30,7 @@ def build_start(kwargs: Dict[str, Any]) -> None:
             raise Exception("start is negative")
 
 
-def _resolve_path_args(kwargs: Dict[str, Any]) -> None:
+def _resolve_path_args(kwargs: Dict[str, Any]) -> Optional[Union[Dict[str, Any], str]]:
     """Resolve the path-type shortcut arguments by precedence.
 
     Pop each key from `kwargs`, even if it's not used.
@@ -55,8 +55,7 @@ def _resolve_path_args(kwargs: Dict[str, Any]) -> None:
             fname = r".*"
         arg = {"$regex": rf"^{dpath}((/)|(/.*/)){fname}$"}
 
-    if arg:
-        kwargs["logical_name"] = arg
+    return arg
 
 
 def build_files_query(kwargs: Dict[str, Any]) -> None:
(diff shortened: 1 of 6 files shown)


ric-evans:
Tests TBD

file_catalog/argbuilder.py
```diff
@@ -31,6 +30,35 @@ def build_start(kwargs: Dict[str, Any]) -> None:
...
+    if "logical_name" in kwargs:
+        arg = kwargs.pop("logical_name")
+
+    # directory & filename
+    if "directory" in kwargs or "filename" in kwargs:
+        if not (dpath := kwargs.pop("directory", "").rstrip("/")):
+            dpath = r".*"
+        if not (fname := kwargs.pop("filename", "").lstrip("/")):
+            fname = r".*"
+        arg = {"$regex": rf"^{dpath}((/)|(/.*/)){fname}$"}
```
dsschult, line 56: I think you need to escape the slashes.

I'm not a regex expert, but wondering if this would be better:
```
^{dpath}\/(.*\/)*{fname}$
```
It has only a single match group.

ric-evans: "`/`"s aren't special characters, at least not in raw strings. `^{dpath}/(.*/)?{fname}$` would work, but I'm not sure it's going to be much of an improvement performance wise

ric-evans: I do like the look of `^{dpath}/(.*/)?{fname}$` so I'm going to use that

dsschult: > "`/`"s aren't special characters, at least not in raw strings. `^{dpath}/(.*/)?{fname}$` would work, but I'm not sure it's going to be much of an improvement performance wise

I know it's not a special character in a string.  Just wondering how mongo is going to parse that.  I gues testing will find out.

ric-evans: ahhh, so far so good

ric-evans: https://stackoverflow.com/a/46616907/13156561

dsschult (approved):

PR #92, opened 2021-03-11
Setup.py Updates & Boiler-plating

ric-evans:


setup.py
```diff
@@ -1,48 +1,84 @@
...
+        "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.3",
+        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
```
dsschult, line 76: Should probably adjust these to match PY_VERSION

setup.py
```diff
@@ -1,48 +1,84 @@
...
-        'Development Status :: 4 - Beta',
-        'License :: OSI Approved :: MIT License',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
+        "Development Status :: 4 - Beta",
```
dsschult, line 66: This could probably be tied to the version number?  I assume anything over 1.0.0 is out of beta.

setup.py
ric-evans: do we need all the python version or just the current one?

ric-evans: I'm including everything from 3.3+. I don't know why we are excluding 3.1-3.2, but I didn't touch that.

dsschult: I would only include the ones we expect it to work on.  So some of the early 3.x probably don't work anymore.

dsschult: Specifically, should python2 be removed now, especially if you start putting in f-strings?

ric-evans: is there a way to test this? Otherwise, I'd just say 3.5

dsschult: Not easily?  You could probably make a bot try out installation/tests on each python version.  But that sounds painful.  You could also do "anything >= PY_VERSION" as a safe bet on what will work.

ric-evans: I added a `OLDEST_PY_VERSION` tuple

setup.py
ric-evans: I added cases for 2-5

setup.py
blinkdog: Yeah, Python 3.6 is minimum for f-strings

dsschult (approved):

PR #101, opened 2021-04-05
remove archive index, as it was getting in the way of a query

dsschult:


dsschult:
I removed this index manually to fix an LTA query (to force it to use the right index).

dsschult:
Question: are the queries in the tests representative of real-world queries?

file_catalog/mongo.py
```diff
@@ -40,7 +40,6 @@ def __init__(self, host=None, port=None, authSource=None, username=None, passwor
         self.client.files.create_index([('logical_name',pymongo.HASHED)], background=True)
         self.client.files.create_index('locations', unique=True, background=True)
         self.client.files.create_index([('locations.site',pymongo.DESCENDING),('locations.path',pymongo.DESCENDING)], background=True)
-        self.client.files.create_index('locations.archive', background=True)
```
ric-evans, line 43: do you think the problem was that locations was double-indexed? One on archive and the other on site/path?

dsschult: Possibly, but the "correct" index for the query in question was on `logical_name`.

dsschult: It could even be that we're running such an old version of Mongo (3.0), that the query planner wasn't smart enough, and just picked the first index it found.

ric-evans: I see. Picking the "right" index would be difficult without prior knowledge I assume

ric-evans: Is there no way of defining an order of precedence for indices?

file_catalog/mongo.py
```diff
@@ -40,7 +40,6 @@ def __init__(self, host=None, port=None, authSource=None, username=None, passwor
         self.client.files.create_index([('logical_name',pymongo.HASHED)], background=True)
```
ric-evans, line 40: also we have two logical_name indexes, what's up with that?

file_catalog/mongo.py
dsschult: Nope, that's not something DBs do.  The best you can do is to "hint" at an index to use when you call a query.  

The DB's query analyzer is supposed to be smart enough to pick the "most optimal" index, but Mongo 3.0 was still relatively young for a DB, unlike MySQL.

ric-evans:
> Question: are the queries in the tests representative of real-world queries?

Which tests? The PR is failing on profiling, but passing the pytests.

file_catalog/mongo.py
dsschult: The HASHED index is a constant time lookup, so if you have the full name it is much faster.  But it doesn't work if you don't have the full name, thus a fallback to the range-based lookup.

dsschult:
> > Question: are the queries in the tests representative of real-world queries?
> 
> Which tests? The PR is failing on profiling, but passing the pytests.

The profiling is run on all tests, if my memory serves.  It just checks that every query hit an index.

blinkdog (approved):
LGTM

ric-evans:
> > > Question: are the queries in the tests representative of real-world queries?
> > 
> > 
> > Which tests? The PR is failing on profiling, but passing the pytests.
> 
> The profiling is run on all tests, if my memory serves. It just checks that every query hit an index.

In that case, this is your problem query: https://github.com/WIPACrepo/file_catalog/blob/master/tests/test_files.py#L362

dsschult:
> > > > Question: are the queries in the tests representative of real-world queries?
> > > 
> > > 
> > > Which tests? The PR is failing on profiling, but passing the pytests.
> > 
> > 
> > The profiling is run on all tests, if my memory serves. It just checks that every query hit an index.
> 
> In that case, this is your problem query: https://github.com/WIPACrepo/file_catalog/blob/master/tests/test_files.py#L362

Yeah, apparently that's not realistic, as we'd never query that generically (we'd get millions of results back).

ric-evans:
it does make me think about whether it would be worth it to throw `archive` into that compound index

ric-evans (approved):
other than the profiling/tests, LGTM

dsschult:
I'm unsure why the profiler is still flagging those queries, since my commits should have fixed that.

ric-evans:
> I'm unsure why the profiler is still flagging those queries, since my commits should have fixed that.

Weird. If you print out the uuids from the POST requests, it could be easier to see what's tripping it up. I see 23 out of 60 POSTs are doing COLLSCAN.

ric-evans:
I'm assuming `db.system.profile.find({ 'op': { '$nin' : ['command', 'insert'] } })` in the profiling code relates to POST

blinkdog:
> I'm unsure why the profiler is still flagging those queries, since my commits should have fixed that.

I think this may be your culprit:
https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/argbuilder.py#L75

It's adding an explicit `query["locations.archive"] = None` to the query by default when `locations.archive` is not specified.
Originally, this would have been "non-member of the archive index", but since the archive index was removed, the only way to know this now is to scan each record.

It changes the API a bit, but I think refactoring that condition and treating it like the others below (if 'archive' is present, then add the explicit value to the query object under the right key) would be a good solution.

@ric-evans @dsschult Thoughts?

ric-evans:
> > I'm unsure why the profiler is still flagging those queries, since my commits should have fixed that.
> 
> I think this may be your culprit:
> [https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/argbuilder.py#L75](https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/argbuilder.py?rgh-link-date=2021-04-07T01%3A07%3A21Z#L75)
> 
> It's adding an explicit `query["locations.archive"] = None` to the query by default when `locations.archive` is not specified.
> Originally, this would have been "non-member of the archive index", but since the archive index was removed, the only way to know this now is to scan each record.
> 
> It changes the API a bit, but I think refactoring that condition and treating it like the others below (if 'archive' is present, then add the explicit value to the query object under the right key) would be a good solution.
> 
> @ric-evans @dsschult Thoughts?

Sure, that makes enough sense to me. This was a some legacy code that I moved over for argbuilder.py--never looked too much at it.

dsschult:
> > I'm unsure why the profiler is still flagging those queries, since my commits should have fixed that.
> 
> I think this may be your culprit:
> https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/argbuilder.py#L75
> 
> It's adding an explicit `query["locations.archive"] = None` to the query by default when `locations.archive` is not specified.
> Originally, this would have been "non-member of the archive index", but since the archive index was removed, the only way to know this now is to scan each record.
> 
> It changes the API a bit, but I think refactoring that condition and treating it like the others below (if 'archive' is present, then add the explicit value to the query object under the right key) would be a good solution.
> 
> @ric-evans @dsschult Thoughts?

So the reason for this condition is to not get archive records by default.  As long as we can maintain that, I'm fine.

ric-evans:
@dsschult Can you expand https://github.com/WIPACrepo/file_catalog/blob/master/tests/test_files.py#L330 if you do remove it? A couple cases with `'archive':False` multiple locations would be useful.

ric-evans (approved):
:+1:

Issue #98, opened 2021-03-30
Remedy `mnt`-Rooted FC Entries That Are Also at NERSC


ric-evans:
(Leftover from https://github.com/WIPACrepo/file_catalog/issues/68)

There are 1679 entries in the FC that are rooted at `/mnt/lfs*/exp` and also duplicated at `/data/exp/`. These files also no longer exist on lfs.

I think we should just delete all 1679x2 FC entries and call it done :smile:

dsschult:
Do either of the entries have NERSC locations?  If so, we need to keep that.

ric-evans:
@jnbellinger By x2 I mean the FC entries rooted at `/mnt/` and those rooted at `/data/`, there are 1679 of each

ric-evans:
@dsschult yes, those rooted at `/mnt/` have NERSC locations. 

But since other entries rooted at `/mnt/` that are no longer in lfs have been deleted from the FC (regardless if they had a NERSC location), I figure deleting these too would be fine. The only difference here is that these 1679 have `/data/`-rooted duplicates.

jnbellinger:
I'm still puzzled.  If we had a file in /data/..., even if we deleted it from the warehouse, so long as it exists in archive we want to keep the info around, right?--just get rid of the duplicate.  Or are there 3 sets of records we're talking about?
Or are  you talking about the dual appearances of PFRawXXX and ukey_HASH_PFRawXXX names for a single file?

ric-evans:
I agree this is getting confusing. I'll run some queries and see if I can clarify things

ric-evans:
I ran queries on each path and now I'm more confused. Regardless, here's the results of the queries:

```
Of the 1679 total paths rooted at /mnt/lfs7/:
- All 1679 are have a NERSC location
- There are 671 with a duplicate rooted at /data/ and those duplicates do not have a NERSC location
- There are 1008 that do not have duplicates
- All 1679 are no longer in lfs/lustre/warehouse
```

And here's my source code for that:

```
In [27]: def has_nersc(file) -> bool: 
    ...:     for loc in file.get('locations',[]): 
    ...:         if loc['site'].upper() == 'NERSC': 
    ...:             return True 
    ...:     return False 
In [28]: import os 
    ...: no_dup = [] 
    ...: has_dup = [] 
    ...: for n in nerscs: 
    ...:     # start w/ the /mnt/-rooted paths 
    ...:     #print(n) 
[...]

PR #102, opened 2021-04-08
Script to Restore From Backup

ric-evans:
Added a script to restore FC entries via their JSON backups.

> Restores FC entries from a JSON file (new-line delimited). The contents of each entry is not altered. However, these keys are removed prior to sending to the File Catalog: ["meta_modify_date", "_id"]. Each entry must include a `uuid` field.

First try to PUT with existing uuid. If that fails (404), then POST with uuid.

You're on your own for any field alterations.


ric-evans:
closes https://github.com/WIPACrepo/file_catalog/issues/98

Issue #103, opened 2021-04-08
404 Does Not Work for PUT & PATCH


ric-evans:
For PUT and PATCH requests, 404s are expected to be sent back if the given uuid is not present in the DB. This is not happening.

This leads to an internal error here, for example: https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/server.py#L619

We should send a 404 ASAP.

PR #104, opened 2021-04-12
Server Spring Cleaning: Code Maintenance & Bug Fixes

ric-evans:
- Continue the slow code maintenance of the File Catalog (_LGTM_ will like this)
- Convert to modern async/await syntax (now using Motor)
- Fix bug where 404s are not sent when a file is not found in PUT and PATCH requests (closes https://github.com/WIPACrepo/file_catalog/issues/103)
- Fix bug where HTTP Error reasons where not passed on to client

### DIFF
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -12,6 +12,8 @@ from typing import Any, Dict, List, Optional, Union, cast
 import pymongo  # type: ignore[import]
 from motor.motor_tornado import MotorClient  # type: ignore[import]
 
+from .schema import types
+
 logger = logging.getLogger("mongo")
 
 
@@ -199,7 +201,7 @@ class Mongo:
 
         return cast(int, ret)
 
-    async def create_file(self, metadata: Dict[str, Any]) -> str:
+    async def create_file(self, metadata: types.Metadata) -> str:
         """Insert file metadata.
 
         Return uuid.
@@ -211,14 +213,14 @@ class Mongo:
             logger.warning(msg)
             raise Exception(msg)
 
-        return cast(str, metadata["uuid"])
+        return metadata["uuid"]
 
-    async def get_file(self, filters: Dict[str, Any]) -> Dict[str, Any]:
+    async def get_file(self, filters: Dict[str, Any]) -> types.Metadata:
         """Get file matching filters."""
         file = await self.client.files.find_one(filters, {"_id": False})
-        return cast(Dict[str, Any], file)
+        return cast(types.Metadata, file)
 
-    async def update_file(self, uuid: str, metadata: Dict[str, Any]) -> None:
+    async def update_file(self, uuid: str, metadata: types.Metadata) -> None:
         """Update file."""
         result = await self.client.files.update_one({"uuid": uuid}, {"$set": metadata})
 
@@ -232,7 +234,7 @@ class Mongo:
             logger.warning(msg)
             raise Exception(msg)
 
-    async def replace_file(self, metadata: Dict[str, Any]) -> None:
+    async def replace_file(self, metadata: types.Metadata) -> None:
         """Replace file.
 
         Metadata must include 'uuid'.
@@ -330,7 +332,10 @@ class Mongo:
         return ret
 
     async def create_snapshot(self, metadata: Dict[str, Any]) -> str:
-        """Insert metadata into 'snapshots' collection."""
+        """Insert metadata into 'snapshots' collection.
+
+        Return uuid.
+        """
         result = await self.client.snapshots.insert_one(metadata)
 
         if (not result) or (not result.inserted_id):
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -62,7 +62,7 @@ class Mongo:
         await self.client.files.create_index('logical_name', unique=True, background=True)
         await self.client.files.create_index([('logical_name', pymongo.HASHED)], background=True)
         await self.client.files.create_index('locations', unique=True, background=True)
-        await self.client.files.create_index([('locations.site', pymongo.DESCENDING), ('locations.path', pymongo.DESCENDING)], background=True,)
+        await self.client.files.create_index([('locations.site', pymongo.DESCENDING), ('locations.path', pymongo.DESCENDING)], background=True)
         await self.client.files.create_index('locations.archive', background=True)
         await self.client.files.create_index('create_date', background=True)
 
@@ -118,29 +118,12 @@ class Mongo:
     async def _limit_result_list(
         cursor: MotorCursor, limit: Optional[int] = None, start: int = 0,
     ) -> List[Dict[str, Any]]:
-        """Get sublist of results from `cursor` using `limit` and `start`.
-
-         `limit` and `skip` are ignored by __getitem__:
-         http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.__getitem__
-
-        Therefore, implement it manually.
-        """
-        results = []
-        end = None
-
-        if limit is not None:
-            end = start + limit
-
-        i = 0
-        async for row in cursor:
-            if i < start:
-                continue
-            results.append(row)
-            if end and i >= end:
-                break
-            i += 1
-
-        return results
+        """Get sublist of results from `cursor` using `limit` and `start`."""
+        if limit:
+            results = await cursor.skip(start).limit(limit).to_list(None)
+        else:
+            results = await cursor.skip(start).to_list(None)
+        return cast(List[Dict[str, Any]], results)
 
     async def find_files(
         self,
--- a/file_catalog/mongo.py
+++ b/file_catalog/mongo.py
@@ -54,68 +54,44 @@ class Mongo:
         self.executor = ThreadPoolExecutor(max_workers=10)
         logger.info("done setting up Mongo")
 
+    # fmt: off
     async def create_indexes(self) -> None:
         """Create indexes for all file-catalog mongo collections."""
         # all files (a.k.a. required fields)
-        await self.client.files.create_index("uuid", unique=True, background=True)
-        await self.client.files.create_index(
-            "logical_name", unique=True, background=True
-        )
-        await self.client.files.create_index(
-            [("logical_name", pymongo.HASHED)], background=True
-        )
-        await self.client.files.create_index("locations", unique=True, background=True)
-        await self.client.files.create_index(
-            [
-                ("locations.site", pymongo.DESCENDING),
-                ("locations.path", pymongo.DESCENDING),
-            ],
-            background=True,
-        )
-        await self.client.files.create_index("locations.archive", background=True)
-        await self.client.files.create_index("create_date", background=True)
+        await self.client.files.create_index('uuid', unique=True, background=True)
+        await self.client.files.create_index('logical_name', unique=True, background=True)
+        await self.client.files.create_index([('logical_name', pymongo.HASHED)], background=True)
+        await self.client.files.create_index('locations', unique=True, background=True)
+        await self.client.files.create_index([('locations.site', pymongo.DESCENDING), ('locations.path', pymongo.DESCENDING)], background=True,)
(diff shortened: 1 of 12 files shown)


ric-evans:
now only 5 tests failing, making progress :sweat_smile:

ric-evans:
I need to add tests for the use cases laid out in the linked issue. Other than that, it's ready for review.

Also fixed a potential bug where `self.assertRaises(Exception)` was being used. The exception that was caught was actually being raised _by the testing code_ (referencing an undeclared variable), but there was no way to tell the difference. So I added a check. That's my PSA.

ric-evans:
@blinkdog can you look at https://github.com/WIPACrepo/file_catalog/blob/patch-put-bug-fixes/file_catalog/server.py#L727 ish? I don't think your note here is accurate:

> note that if we get the record that we are trying to update
 the location will NOT be added to the list of new_locations
 which leaves new_locations as a vetted list of addable locations

ric-evans:
@dsschult @blinkdog Okay, I've finished everything I planned to do, plus all the incidental things (incidental >> planned). It's ready for review.

ric-evans:
@dsschult for many of the `send_error()` calls, we're including `file` and/or `location`. Looking at `APIHandler.write_error()` (https://github.com/WIPACrepo/file_catalog/blob/patch-put-bug-fixes/file_catalog/server.py#L396), these args are written to the output buffer. But I don't see how that's visible to the client. And in my tests, I haven't seen these show up in the request's http errors.

This is also why I refactored `message` to `reason`. `reason` is visible to the client. So no more bland `"Bad Request"`. Not sure if this was the intended behavior and never got addressed... but it sure was a rabbit hole :smile:

dsschult:
> @dsschult for many of the `send_error()` calls, we're including `file` and/or `location`. Looking at `APIHandler.write_error()` (https://github.com/WIPACrepo/file_catalog/blob/patch-put-bug-fixes/file_catalog/server.py#L396), these args are written to the output buffer. But I don't see how that's visible to the client. And in my tests, I haven't seen these show up in the request's http errors.
> 
> This is also why I refactored `message` to `reason`. `reason` is visible to the client. So no more bland `"Bad Request"`. Not sure if this was the intended behavior and never got addressed... but it sure was a rabbit hole smile

I think this was intended to log the message, but not output it back to the client.  I know it's bad practice to output unsanitized input back to the client,

ric-evans:
> > @dsschult for many of the `send_error()` calls, we're including `file` and/or `location`. Looking at `APIHandler.write_error()` ([https://github.com/WIPACrepo/file_catalog/blob/patch-put-bug-fixes/file_catalog/server.py#L396](https://github.com/WIPACrepo/file_catalog/blob/patch-put-bug-fixes/file_catalog/server.py?rgh-link-date=2021-04-14T23%3A52%3A46Z#L396)), these args are written to the output buffer. But I don't see how that's visible to the client. And in my tests, I haven't seen these show up in the request's http errors.
> > This is also why I refactored `message` to `reason`. `reason` is visible to the client. So no more bland `"Bad Request"`. Not sure if this was the intended behavior and never got addressed... but it sure was a rabbit hole smile
> 
> I think this was intended to log the message, but not output it back to the client. I know it's bad practice to output unsanitized input back to the client,

Alright, that's what I was thinking. I haven't removed any of those "extra" arguments. So, it's still backwards compatible, aside from `message`/`reason`.

blinkdog (approved):
Only one thing to possibly change, although it might be out of scope; there is quite a bit here already!

Anyway, LGTM

file_catalog/mongo.py
```diff
@@ -247,18 +264,17 @@ def find_collections(
...
-        ret = Mongo._limit_result_list(result, limit, start)
+        cursor = self.client.collections.find({}, projection)
+        results = await Mongo._limit_result_list(cursor, limit, start)
 
-        return ret
+        return results
 
-    @run_on_executor
-    def create_collection(self, metadata: Dict[str, Any]) -> str:
+    async def create_collection(self, metadata: Dict[str, Any]) -> str:
```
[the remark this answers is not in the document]
ric-evans: Right, it's a "collection"'s metadata. Not to be confused with a "file"'s "Metadata", or a mongo collection of "files"... sigh.

I tried to stay clear of "collection" and "snapshot" things.

file_catalog/server.py
```diff
@@ -158,54 +166,69 @@ def __init__(self, config, port=8888, debug=False,
...
-    def initialize(self, base_url='/', debug=False, config=None):
+
+    def initialize(  # pylint: disable=C0116,W0201
+        self,
+        base_url: str = "/",
+        debug: bool = False,
+        config: Optional[Dict[str, Any]] = None,
+    ) -> None:  # noqa: D102
+        if config is None:
+            raise Exception('Invalid configuration: `config` is `None`')
```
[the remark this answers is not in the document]
ric-evans: Good idea. Since these args are passed via unpacking kwargs in `tornado.web.RequestHandler`, the signature's ordering doesn't matter here.

ric-evans:
@blinkdog One thing I'd like your input about is the several mypy errors (`mypy: error
override - Signature of "initialize" incompatible with supertype "APIHandler"`) for the derived class' `initialize()` methods. Since we're using `kwargs` and immediately calling `super().initialize(**kwargs)` it's not _terrible_. But mypy disagrees.

blinkdog:
> @blinkdog One thing I'd like your input about is the several mypy errors (`mypy: error override - Signature of "initialize" incompatible with supertype "APIHandler"`) for the derived class' `initialize()` methods. Since we're using `kwargs` and immediately calling `super().initialize(**kwargs)` it's not _terrible_. But mypy disagrees.

So the superclass of `APIHandler` (one of ours) is `RequestHandler` (from Tornado) and the docs give it this signature:
https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.initialize

We can use `kwargs`, but I think replicating the signature down is probably the right way to go.
The tests confirm that it's not a correctness issue, but I think taking mypy's advice is a good idea.

dsschult:
> > @blinkdog One thing I'd like your input about is the several mypy errors (`mypy: error override - Signature of "initialize" incompatible with supertype "APIHandler"`) for the derived class' `initialize()` methods. Since we're using `kwargs` and immediately calling `super().initialize(**kwargs)` it's not _terrible_. But mypy disagrees.
> 
> So the superclass of `APIHandler` (one of ours) is `RequestHandler` (from Tornado) and the docs give it this signature:
> https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.initialize
> 
> We can use `kwargs`, but I think replicating the signature down is probably the right way to go.
> The tests confirm that it's not a correctness issue, but I think taking mypy's advice is a good idea.

I'll disagree with @blinkdog about removing kwargs.  Both `*args` and `**kwargs` are fundamental parts of Python, in that you can "pass along" arguments without explicitly knowing what they are.  This is very handy when creating a wrapper around something, since the underlying object can change its arguments and your library doesn't need to be updated (even if your users have to update their code).

I also believe this is a step too far in pursuing fully typed Python - we might as well switch to a fully typed language like Rust then :smile:, as I think we start to lose some of the strengths of Python.

ric-evans:
> > > @blinkdog One thing I'd like your input about is the several mypy errors (`mypy: error override - Signature of "initialize" incompatible with supertype "APIHandler"`) for the derived class' `initialize()` methods. Since we're using `kwargs` and immediately calling `super().initialize(**kwargs)` it's not _terrible_. But mypy disagrees.
> > 
> > 
> > So the superclass of `APIHandler` (one of ours) is `RequestHandler` (from Tornado) and the docs give it this signature:
> > [tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.initialize](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.initialize)
> > We can use `kwargs`, but I think replicating the signature down is probably the right way to go.
> > The tests confirm that it's not a correctness issue, but I think taking mypy's advice is a good idea.
> 
> I'll disagree with @blinkdog about removing kwargs. Both `*args` and `**kwargs` are fundamental parts of Python, in that you can "pass along" arguments without explicitly knowing what they are. This is very handy when creating a wrapper around something, since the underlying object can change its arguments and your library doesn't need to be updated (even if your users have to update their code).
> 
> I also believe this is a step too far in pursuing fully typed Python - we might as well switch to a fully typed language like Rust then , as I think we start to lose some of the strengths of Python.

I do find it surprising that mypy can't track a parameter via kwargs to its parent class.

ric-evans:
I will add, in support of type-hinting, running `mypy` will only complain if there are errors. So any legacy/non-typed code is essentially ignored.

Locally, I run with the `--strict` flag, but that's overkill for our CI needs.

file_catalog/mongo.py
```diff
@@ -19,57 +22,76 @@ class AllKeys:  # pylint: disable=R0903
...
+
+    # fmt: off
+    async def create_indexes(self) -> None:
+        """Create indexes for all file-catalog mongo collections."""
+        # all files (a.k.a. required fields)
+        await self.client.files.create_index('uuid', unique=True, background=True)
+        await self.client.files.create_index('logical_name', unique=True, background=True)
+        await self.client.files.create_index([('logical_name', pymongo.HASHED)], background=True)
+        await self.client.files.create_index('locations', unique=True, background=True)
+        await self.client.files.create_index([('locations.site', pymongo.DESCENDING), ('locations.path', pymongo.DESCENDING)], background=True,)
```
dsschult, line 65: Style nitpick: extra comma in arg list.

file_catalog/mongo.py
```diff
@@ -93,29 +115,34 @@ def _get_projection(
...
+        results = []
         end = None
 
         if limit is not None:
             end = start + limit
 
-        for row in result[start:end]:
-            ret.append(row)
+        i = 0
+        async for row in cursor:
```
dsschult, line 135: It might be better to use ` results = cursor.skip(start).limit(limit).to_list(None)`.  This lets it use server-side skipping.

dsschult: It's also short enough it might remove the need for this function :smile:

file_catalog/pathfinder.py
```diff
@@ -0,0 +1,81 @@
...
+        # try to load a file by that logical_name
+        file_found = await apihandler.db.get_file(
+            {"logical_name": metadata["logical_name"]}
+        )
+        # if we got a file by that logical_name
+        if _is_conflict(uuid, file_found):
+            # then that logical_name belongs to another file (already exists)
+            apihandler.send_error(
+                409,
+                reason=f"Conflict with existing file (logical_name already exists `{metadata['logical_name']}`)",
```
dsschult, line 36: From a security perspective, returning the source metadata without sanitizing it in the reason isn't great.  It might be better to just say:
```python
reason="Conflict with existing file (logical_name already exists)"
```

file_catalog/pathfinder.py
```diff
@@ -0,0 +1,81 @@
...
+            # try to load a file by that location
+            file_found = await apihandler.db.get_file(
+                {"locations": {"$elemMatch": loc}}
+            )
+            # if we got a file by that location
+            if _is_conflict(uuid, file_found):
+                # then that location belongs to another file (already exists)
+                apihandler.send_error(
+                    409,
+                    reason=f"Conflict with existing file (location already exists `{loc['path']}`)",
```
dsschult, line 59: See above comment about unsanitized input.

dsschult (approved):

file_catalog/mongo.py
ric-evans: good catch

file_catalog/mongo.py
ric-evans: Sure, that'd be much better. What's up with `to_list(None)`? Why `None`?

dsschult: `to_list` takes a limit parameter, for how many items it should consume before stopping.  `None` means infinite.

I suppose for a production instance you'd probably impose some upper bound for performance reasons, and if you hit that you'd want to return an error.  But since I think we're keeping FC only for in-house use, it's fine to use infinite.

PR #107, opened 2021-05-27
Setup Updates: SetupShop and GH Action

ric-evans:


ric-evans:
@dsschult is that Docker Action supposed to be running in a PR?

ric-evans:
there's a bunch of dependency conflicts with rest-tools, which is a "good" thing. Planning to bump a lot of reqs in `requirements.txt`

dsschult:
> @dsschult is that Docker Action supposed to be running in a PR?

You mean the docker one?  Yes.  In a PR it just builds, but doesn't push.

ric-evans:
> > @dsschult is that Docker Action supposed to be running in a PR?
> 
> You mean the docker one? Yes. In a PR it just builds, but doesn't push.

Got it. The issue looks to be the same as the other tests

PR #108, opened 2021-07-06
Added LTA type hints

blinkdog:
Per suggestion by @ric-evans last week:
https://github.com/WIPACrepo/lta/pull/205#issuecomment-870686163

Does this look good Ric?


ric-evans (approved):
Looks good. Someday it'd be nice to incorporate all these definitions into a validation step.

Issue #109, opened 2021-09-09
make syncing script between fs and FC


dsschult:
The syncing script should do more than the current indexing script:

1. index new files
2. check timestamp/checksum of existing files. re-index any changed files.
    a. If FC entry has a location not at WIPAC, change logical_name and remove WIPAC location.
3. "remove" location of files that have been deleted/overwritten
	a. "remove" could mean outright deleting the entry, or adding a flag to the entry

ric-evans:
Note: This would be ran manually, ad-hoc, immediately after a known bulk removal is made.

ric-evans:
An issue with deleting a location is that the `logical_name` would already exist for a re-written/restored file.

A solution would be to no longer enforce unique `logical_name`s, and add logic to handle this scenario in the Indexing script / FC REST interface. Currently, you cannot add a record with the same `logical_name`, unless you PATCH (in which case you would lose the previous metadata).

ric-evans:
### Solution I: Non-Unique `logical_name`
A solution would be to no longer enforce unique `logical_name`s, and add logic to handle this scenario in the Indexing script / FC REST interface. Currently, you cannot add a record with the same `logical_name`. You can PATCH, but you would lose the previous metadata... not good.
- Example Future Scenario:
  + I want the FC record at `/my/data/warehouse/path/to/file`
  + So, I call `/api/files` with query, `{'logical_name':'/my/data/warehouse/path/to/file'}`
  + FC returns two files with `logical_name` = `/my/data/warehouse/path/to/file`
  + Do I use record A or record B? 
     - My existing script probably assumes there's only one matching file here (`my_file = ret[0]`)

ric-evans:
### Solution II: Prepend/Append Metadata to `logical_name`

When a file is removed from lfs (but still exists at NERSC), append `"@nersc"` to the `logcal_name`.
- Example Future Scenario 1:
  + I want the FC record at `/my/data/warehouse/path/to/file`
  + So, I call `/api/files` with query, `{'logical_name':'/my/data/warehouse/path/to/file'}`
  + FC returns **one** file with `logical_name` = `/my/data/warehouse/path/to/file`
  + All is good, I got what I wanted :smile: 
- Example Future Scenario 2:
  + I want all FC records at `/my/data/warehouse/path/to/parent/directory`
  + So, I call `/api/files` with query, `{'directory':'/my/data/warehouse/path/to/parent/directory'}`
[...]

PR #110, opened 2021-09-20
Allow Non-Unique `logical_name`s

ric-evans:
- Allows the use of non-unique logical names
	+ This allows the versioning of files. For example, version 0 is at NERSC and version 1 is at WIPAC.
	+ `checksum`+`logical_name` is unique. This represents a file version. The `create_date` identifies the age of the file version.
- Validates incoming location-type entries

from https://github.com/WIPACrepo/file_catalog/issues/109

### DIFF
--- a/file_catalog/schema/validation.py
+++ b/file_catalog/schema/validation.py
@@ -28,6 +28,12 @@ def _find_missing_mandatory_field(metadata: types.Metadata, fields: List[str]) -
 class Validation:
     """Validating field-specific metadata."""
 
+    MANDATORY_LOCATION_KEYS = ['site', 'path']
+    INVALID_LOCATIONS_LIST_MESSAGE = (
+        f"Validation Error: member `locations` must be a list with "
+        f"1+ entries, each with keys: {MANDATORY_LOCATION_KEYS}"
+    )
+
     def __init__(self, config: Dict[str, Any]) -> None:
         self.config = config
 
@@ -35,6 +41,33 @@ class Validation:
     def is_valid_sha512(hash_str: str) -> bool:
         """Check if `hash_str` is a valid SHA512 hash."""
         return re.match(r"[0-9a-f]{128}", str(hash_str), re.IGNORECASE) is not None
+
+    @staticmethod
+    def is_valid_location_list(locations: List[types.LocationEntry]) -> bool:
+        """Check if `locations` is a valid list of location-entries."""
+        if not isinstance(locations, list):
+            return False
+        if not locations:
+            return False
+
+        for loc in locations:
+            if not Validation.is_valid_location(loc):
+                return False
+
+        return True
+
+    @staticmethod
+    def is_valid_location(location: types.LocationEntry) -> bool:
+        """Check if `location` is a valid location-entry."""
+        if not location:
+            return False
+        if not isinstance(location, dict):
+            return False
+
+        if not all(key in location for key in Validation.MANDATORY_LOCATION_KEYS):
+            return False
+
+        return True
 
     def has_forbidden_attributes_creation(self, apihandler: Any, metadata: types.Metadata, old_metadata: types.Metadata) -> bool:
         """Check if `metadata` has forbidden attributes and they have changed.
@@ -75,6 +108,7 @@ class Validation:
         Utilizes `send_error` and returns `False` if validation failed.
         If validation was successful, `True` is returned.
         """
+        # MANDATORY FIELDS
         missing = _find_missing_mandatory_field(metadata, self.config['META_MANDATORY_FIELDS'])
         if missing:
             apihandler.send_error(
@@ -85,6 +119,7 @@ class Validation:
             )
             return False
 
+        # CHECKSUM
         if ((not isinstance(metadata['checksum'], dict)) or 'sha512' not in metadata['checksum']):
             # checksum needs to be a dict with an sha512
             apihandler.send_error(
@@ -94,7 +129,8 @@ class Validation:
             )
             return False
 
-        elif not self.is_valid_sha512(metadata['checksum']['sha512']):
+        # CHECKSSUM.SHA512
+        if not self.is_valid_sha512(metadata['checksum']['sha512']):
             # force to use SHA512
             apihandler.send_error(
                 400,
@@ -103,13 +139,11 @@ class Validation:
             )
             return False
 
-        elif ((not isinstance(metadata['locations'], list))
-              or (not metadata['locations'])
-              or not all(loc for loc in metadata['locations'])):
-            # locations needs to be a non-empty list
+        # LOCATIONS LIST & ITS ENTRIES
+        if not self.is_valid_location_list(metadata['locations']):
             apihandler.send_error(
                 400,
-                reason='Validation Error: member `locations` must be a list with at least one entry',
+                reason=self.INVALID_LOCATIONS_LIST_MESSAGE,
                 file=apihandler.files_url
             )
             return False
--- a/file_catalog/server.py
+++ b/file_catalog/server.py
@@ -348,6 +348,8 @@ class APIHandler(tornado.web.RequestHandler):
         self.base_url = base_url
         self.debug = debug
         self.config = config
+        self.validation = Validation(self.config)
+
         if 'TOKEN_KEY' in self.config:
             self.auth = Auth(algorithm=self.config['TOKEN_ALGORITHM'],
                              secret=self.config['TOKEN_KEY'],
@@ -437,7 +439,6 @@ class FilesHandler(APIHandler):
         super().initialize(**kwargs)
         # pylint: disable=W0201
         self.files_url = os.path.join(self.base_url, 'files')
-        self.validation = Validation(self.config)  # pylint: disable=W0201
 
     @validate_auth
     @catch_error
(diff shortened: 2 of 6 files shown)


dsschult (commented):
Generally looks fine, just need to resolve the failing tests.

ric-evans:
@dsschult I didn't have the bulk of the added/refined logic when you last looked at this

ric-evans:
I still need to write/fix tests

ric-evans:
@dsschult @blinkdog - I distinguish two types of guardrails in the various REST methods: validation (static/stateless) and deconfliction (stateful). Validation checks the schema of the incoming data, whereas Deconfliction checks whether the incoming data will conflict with something already in the databases.

ric-evans:
Old tests are fixed and/or deleted (where logic changed drastically). New tests tomorrow.

file_catalog/server.py
```diff
@@ -474,43 +476,39 @@ async def post(self) -> None:
...
 
-                await self.db.update_file(db_file['uuid'], {'locations': db_file['locations']})
-                self.set_status(200)
-                uuid = db_file['uuid']
-        else:
-            uuid = await self.db.create_file(metadata)
-            self.set_status(201)
+        # Create & Write-Back
+        await self.db.create_file(metadata)
+        self.set_status(201)
```
blinkdog, line 505: I see we lost the "additional replica" of this POST route, but I'm pretty sure we talked about how this route would handle new records and `/locations` would handle adding locations.

file_catalog/server.py
```diff
@@ -651,8 +656,17 @@ async def put(self, uuid: str) -> None:
...
+        # Deconflict with DB Records
+        # NOTE - PUT should not conflict with any existing record (excl. uuid's record)
+        # NOTE - by existing location(s) or by existing file-version
+        if await deconfliction.any_location_in_db(self, metadata.get("locations"), skip=uuid):
             return
+        try:
+            if await deconfliction.FileVersion(metadata).is_in_db(self, skip=uuid):
+                return
+        except deconfliction.IndeterminateFileVersionError:
+            pass
```
blinkdog, line 669: Do we want to pass here?

I understand passing in the PATCH case, because we may not provide checksum and/or logical_name in a PATCH body, depending on what we're patching.

The PUT route is 'replace the resource' and a valid replacement shouldn't be throwing us an IndeterminateFileVersionError due to one of those fields being missing?

blinkdog (changes requested):
Overall it looks good. I had some minor comments that might be worth investigating.

file_catalog/schema/validation.py
```diff
@@ -36,26 +49,81 @@ def is_valid_sha512(hash_str: str) -> bool:
         """Check if `hash_str` is a valid SHA512 hash."""
         return re.match(r"[0-9a-f]{128}", str(hash_str), re.IGNORECASE) is not None
 
+    @staticmethod
+    def is_valid_location_list(locations: List[types.LocationEntry]) -> bool:
+        """Check if `locations` is a valid list of location-entries."""
+        if not isinstance(locations, list):
+            return False
+        if not locations:
+            return False
```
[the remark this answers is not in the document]
ric-evans: In the second `if`, I'm looking to see if the list is empty.

file_catalog/server.py
ric-evans: Can you elaborate on this?

file_catalog/schema/validation.py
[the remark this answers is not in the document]
dsschult: Technically the order doesn't matter, because a `None` will still work for `isinstance`.  But maybe it's even better to combine the two?

```python
if (not locations) or not isinstance(locations, list):
    return False
```

file_catalog/server.py
dsschult: I agree with @blinkdog

dsschult (commented):
Other than the comments raised by @blinkdog it looks good to me.

file_catalog/server.py
ric-evans: Great catch. I've re-shuffled the validation & deconfliction logic in a new commit. I also replaced `pass` with an error for good measure.

file_catalog/schema/validation.py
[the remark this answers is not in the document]
blinkdog: Ahhh, I see. I was assuming it was like the input validation in `is_valid_location` below.
Considering `isinstance` does the work of null checking, then this is fine.

I prefer keeping it separate rather than combining it.
Same logic though, so it's just preference, not correctness. :shrug:

file_catalog/server.py
blinkdog: On old line 486, we start two paths:
- If we found a file by that UUID, we check to see if we're breaking something (409), or if we add a new 'locations' entry `# add replica` (200)

- If we didn't find a file by that UUID, we create a new record (201)


The new code checks to make sure the POST body is
- not providing anything that shouldn't be provided (i,e.: mongo_id, meta_modify_date)
- providing everything that should be provided (i.e.: logical_name, locations, checksum, etc.)
- not overwriting an existing uuid
- not matching an existing location
- not matching an existing file-version (logical_name, checksum)

Once it passes the gauntlet, the new metadata record is created (201).


Functionally, this POST route lost that `# add replica` (200):
- there is no way to get an additional location in an existing record and a 200 code out of this route now
 
However, I think we discussed that, and how LTA was using the `.../{uuid}/locations` route exclusively for adding new locations. So that add replica functionality wasn't used in this route and isn't necessary.

Not suggesting any code change here; I think it looks good. I just wanted to confirm that's the change we were talking about and make sure it's intentional (I think it is?)

dsschult: I confirm that's the intended change.

file_catalog/schema/validation.py
```diff
@@ -88,57 +89,69 @@ def _find_all_field_vals(metadata: types.Metadata, fields: List[str]) -> Dict[st
...
+    def has_forbidden_attributes_modification(
+        self, apihandler: Any, metadata: types.Metadata, old_metadata: types.Metadata
+    ) -> bool:
+        """Check if `metadata` has forbidden attribute updates."""
+        return self._has_forbidden_attributes(
+            apihandler,
+            metadata,
+            old_metadata,
+            self.FORBIDDEN_FIELDS_MODIFICATION,
+            "forbidden attribute update",
```
blinkdog, line 148: You changed the name from update to modification, did you want to change this string too?

file_catalog/schema/validation.py
```diff
@@ -88,57 +89,69 @@ def _find_all_field_vals(metadata: types.Metadata, fields: List[str]) -> Dict[st
...
+                    reason=f"Validation Error: {http_error_message} '{field}'",
                     file=apihandler.files_url,
                 )
                 return True
         return False
 
-    def validate_metadata_creation(self, apihandler: Any, metadata: types.Metadata) -> bool:
-        """Validate metadata for creation.
+    def has_forbidden_attributes_creation(
+        self, apihandler: Any, metadata: types.Metadata, old_metadata: types.Metadata
```
blinkdog, line 125: I think this signature is consistent with `has_forbidden_attributes_modification` below, but does it need `old_metadata` in the case of creation?

On line 480 below, you are passing in `{}`, as an empty object to compare to (which makes sense, in the case of creation).

I think you might be able to remove `old_metadata` as a parameter and just define it as a local variable in the function like: `old_metadata = {}` between line 130 and 131.

blinkdog (changes requested):
These comments are nit-picking. If you think they aren't necessary changes, feel free to push back and I'll change it to Approve rather than Request changes. :smile:

ric-evans:
> These comments are nit-picking. If you think they aren't necessary changes, feel free to push back and I'll change it to Approve rather than Request changes.

it does help that we're nitpicky about similar things :smile:

file_catalog/schema/validation.py
ric-evans: sure

file_catalog/schema/validation.py
ric-evans: I'm glad you caught this. I'd been wondering about this myself, most likely it was an artifact of some refactoring.

blinkdog (approved):
LTGM

ric-evans:
we'll need to change the MongoDB `logical_name` index to `unique=False` in the active db

ric-evans:
wait for https://github.com/WIPACrepo/wipac-telemetry-prototype/issues/30 before merge, then require the newest WIPACTEL version

ric-evans:
A few more things need to happen before merge:
- manually update the `logical_name` index on the current MongoDB
- https://github.com/WIPACrepo/file-catalog-indexer/issues/31
- https://github.com/WIPACrepo/lta/issues/208

dsschult:
>     * manually update the `logical_name` index on the current MongoDB

We should do this in the hours before the new version is pushed to prod, so the old version doesn't restart and try re-creating the old index.  Note that I think this needs to be done manually because the index is named the same, and it won't know to delete and re-create the index.

ric-evans:
@blinkdog @dsschult looking to merge this in on Oct 6. After the client scripts have been taken care of.

ric-evans:
closes  #113

PR #112, opened 2021-09-29
Add Timeout to Mongo Queries

ric-evans:
This will solve the problem of bad queries hanging the server.

1. Add a 10-minutes default timeout to all Mongo queries
2. Add a argument to `/api/files`:`GET` called `max_time_ms` to configure this timeout

merge after https://github.com/WIPACrepo/file_catalog/pull/110


ric-evans:
FYI I branched from https://github.com/WIPACrepo/file_catalog/pull/110 hence all the commits, only the last two are relevant.

ric-evans:
replaced with https://github.com/WIPACrepo/file_catalog/pull/114

PR #116, opened 2021-10-05
Add "Remove Location" Route

ric-evans:
Adds `/api/files/{uuid}/actions/remove_location`:`POST` with a single `location` argument, and required keys, `site` and `path`, which does the following:
1. Removes the `locations` entry
2. If the record now has no `locations` entries, silently delete the record

closes https://github.com/WIPACrepo/file_catalog/issues/109


ric-evans:
All these commits are from branching from https://github.com/WIPACrepo/file_catalog/pull/110.

**Actual commits start here:**

blinkdog (approved):
LGTM.

At first, I was a little conflicted about throwing a 404 for two different reasons:

1. I thought there was a better HTTP status code for a missing item, maybe like 400 for a bad request. Looking into it though, 400 seems to be more for "sorry, this isn't JSON, so I have no idea how to interpret it" rather than "you specified something that doesn't exist". The latter case IS appropriately covered by the 404 status code.

2. I wondered if the action should act like a DELETE verb and just return a "yep, that resource doesn't exist" response (OK) instead of throwing an error. However, this IS a POST verb and there are no guarantees that it will be idempotent, so expecting it to act like a DELETE probably isn't right.

Finally, I settled into a practical outcome: If we are deleting locations that don't exist, that's probably an error we'd like to know about. Most likely it means we're NOT deleting a location that we do want to delete, save for some bug we don't intend.

So yeah, all that said, this looks good.

file_catalog/server.py
```diff
@@ -686,6 +687,101 @@ async def put(self, uuid: str) -> None:
...
+            self.send_error(
+                404,
+                reason=f"Location entry not found for site='{site}' & path='{path}'"
+            )
+            return
+        # remove location! -- there are remaining locations after filtering
+        elif after:
+            await self.db.update_file(uuid, {'locations': after})
+            # re-read the updated record from the database
+            db_file = await self.db.get_file({'uuid': uuid})
```
dsschult, line 762: Consider either modifying or creating a new db method that does `find_one_and_update` so it can atomically update the file and return it.

Issue #118, opened 2021-10-07
add tracing


dsschult:
- db layer spans
- anything else?

then deploy and point at our beta tracing instance

PR #121, opened 2021-11-02
Add DB Tracing

ric-evans:
In addition to built-in `rest-tools` tracing, trace everything that talks to Mongo.

closes https://github.com/WIPACrepo/file_catalog/issues/118


ric-evans:
@blinkdog Open question: Is there such a thing as too much tracing?

I've traced things that _can_ take significant time, like db calls. I've set `all_arg=True` for all the traces, I figure this does no harm and is quicker than manually deciding what's worth/ not worth attributing.

blinkdog (approved):
LGTM. I may need to crib the `all_args=True` for the LTA tracing, :thinking:

ric-evans:
> LGTM. I may need to crib the `all_args=True` for the LTA tracing, thinking

It's handy for sure. The only reason it's not on by default is for the case where we have sensitive data in the args.

Issue #122, opened 2021-11-03
Use `rest_tools.server.RestHandler`


ric-evans:
The File Catalog isn't using `rest_tools.server.RestHandler` so we don't have built-in client-server tracing.

The stuff in https://github.com/WIPACrepo/file_catalog/blob/master/file_catalog/server.py#L335 will need a thorough review, though most of it looks like it can be deleted.

PR #124, opened 2021-11-10
Use `rest_tools.server.RestHandler` for Full Tracing

ric-evans:
Adds full WIPACTEL instrumentation

closes https://github.com/WIPACrepo/file_catalog/issues/122


ric-evans:
This is just a bit more than the minimum necessary to get auto-instrumented traces with the FC.

It doesn't use `rest_tools.handler.RestHandlerSetup`, that would take a lot more work than is necessary. It's also not using the auth wrappers: `@scope_role_auth()` or `@role_authorization()` (@dsschult, feel free to look at that if you want).

file_catalog/server.py
```diff
@@ -332,78 +322,32 @@ def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
     return wrapper
```
dsschult, line 322: I've checked out the differences, and I think this block can be straight deleted and we can use the
```python
@handler.authenticated
```
decorator.  The main thing the current version has is a pass-through when auth is disabled.  I'm not sure if that would break a test, but production would be fine.

dsschult (approved):

blinkdog (commented):
Nothing bad jumps out at me.

I didn't look at the `@handler.authenticated` decorator like David did, but if it works it seems sensible to re-use it and remove the file_catalog specific one if possible.

I notice in rest-tools v1.1.23 that mypy doesn't really like the way we use types; there are more comments to override type 'errors' than before. Not sure if that's mypy being too picky or us being too loose / going against the grain / etc. Not that I'm volunteering to re-design the RestHandler API! :sweat_smile: Just thinking about the health of our canary as we descend into the coal mine.

ric-evans:
> Nothing bad jumps out at me.
> 
> I didn't look at the `@handler.authenticated` decorator like David did, but if it works it seems sensible to re-use it and remove the file_catalog specific one if possible.
> 
> I notice in rest-tools v1.1.23 that mypy doesn't really like the way we use types; there are more comments to override type 'errors' than before. Not sure if that's mypy being too picky or us being too loose / going against the grain / etc. Not that I'm volunteering to re-design the RestHandler API! sweat_smile Just thinking about the health of our canary as we descend into the coal mine.

@blinkdog I saw this in your comment on https://github.com/WIPACrepo/lta/issues/216. What are the actual mypy complaints/errors/warnings you're seeing? The CircleCI mypy job looks fine.

.circleci/config.yml
```diff
@@ -34,9 +35,10 @@ jobs:
         steps:
             - checkout
             - run: |
-                ./setupenv.sh &&
-                . env/bin/activate &&
-                pip install -r requirements.txt -r requirements-dev.txt &&
+                ./setupenv.sh
+                . env/bin/activate
+                pip install -r requirements.txt
```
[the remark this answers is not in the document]
ric-evans: good catch, we don't

file_catalog/server.py
ric-evans: `@handler.authenticated` is much simpler. Is this because `self.current_user`/`self.get_current_user()` is doing the actual auth?

dsschult: Yes, that's where much of the code went.

Issue #127, opened 2022-03-24
Publish to PyPI


ric-evans:
See https://github.com/WIPACrepo/wipac-dev-tools/issues/20

This repo is treated as a package by the Indexer (mainly `file_catalog.schema`).

PR #128, opened 2022-04-08
Publish to PyPI

ric-evans:
closes https://github.com/WIPACrepo/file_catalog/issues/127


ric-evans:
prev commit (2 back) should read "bump `WIPACrepo/wipac-dev-py-setup-action`"

dsschult (approved):
I think this is fine.  We'd have to deploy it to really be sure.

PR #138, opened 2023-02-02
Keycloak Authentication

### LINKED ISSUE #111: Replace `write_error`-`return` with `raise`-`HTTPError` Syntax
remembering to return after writing the error is prone to create bugs, especially when the writing is nested in a call stack.

Instead, simply `raise HTTPError`. Some refactoring will be needed for validation/deconfliction methods that return bools.

### LINKED ISSUE #135: Keycloak Integration for File Catalog

blinkdog:
### File Catalog Auth Update

- File Catalog now requires at least Python 3.8
- Migrated CircleCI `config.yml` to GitHub Actions `profile_queries.yml`
- Updated `wipac-cicd.yml`
- Cleaned up CircleCI and Vagrant references in `README.md`
- Modified environment variables listed in `file-catalog.sh` script
- Removed `__future__` imports; the future is now!
- Fixed flake8 and mypy issues across the codebase
- Modified `__main__.py` to use asyncio.run loop
- ***Modified `server.py` to use new keycloak_role_auth*** <-- Hey, I knew there was auth in here somewhere!
- Removed web interface stubs from `server.py`
- Rewrote urlargparse to handle parsing complex jQuery style URL args
- Modified route handlers to use 'raise HTTPError' instead of 'send_error'
- Removed catch handlers for `pymongo.errors.InvalidId` (removed from PyMongo)
- Moved dev requirements to setup.cfg; bot now maintains `requirements-dev.txt`
- Added `suru` script as a development helper
- Created PyTest fixtures in `conftest.py`
- Migrated unit tests to async-compatible PyTest unit tests

Closes #111 
Closes #135


.github/workflows/wipac-cicd.yml
```diff
@@ -12,25 +12,39 @@ jobs:
...
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
-      - uses: WIPACrepo/wipac-dev-py-setup-action@v1.4
+      - if: |
+          github.actor != 'dependabot[bot]' &&
+          github.ref_type == 'branch' &&
+          format('refs/heads/{0}', github.event.repository.default_branch) != github.ref
+        name: wipac-dev-py-setup-action (only for non-dependabot non-default branches)
+        uses: WIPACrepo/wipac-dev-py-setup-action@v1.9
```
ric-evans, line 45: `v1.10` is out now

blinkdog: Fixed. Thanks! :smile:

blinkdog:
The bot updated requirements.txt and pinned the conflicted requirements to specific versions.
`suru check` may suggest upgrades that can't really be done.

    googleapis-common-protos can be upgraded from 1.56.2 to 1.58.0
    protobuf can be upgraded from 3.20.3 to 4.21.12
    pyjwt can be upgraded from 2.5.0 to 2.6.0

ric-evans:
> The bot updated requirements.txt and pinned the conflicted requirements to specific versions. `suru check` may suggest upgrades that can't really be done.
> 
> ```
> googleapis-common-protos can be upgraded from 1.56.2 to 1.58.0
> protobuf can be upgraded from 3.20.3 to 4.21.12
> pyjwt can be upgraded from 2.5.0 to 2.6.0
> ```

ack! `protobuf` has had conflicting version bumps for a while now... it wouldn't be the worse thing to pin these in setup.cfg. We'll just have to remember to bump them later-- O(months)

dsschult:
> > The bot updated requirements.txt and pinned the conflicted requirements to specific versions. `suru check` may suggest upgrades that can't really be done.
> > ```
> > googleapis-common-protos can be upgraded from 1.56.2 to 1.58.0
> > protobuf can be upgraded from 3.20.3 to 4.21.12
> > pyjwt can be upgraded from 2.5.0 to 2.6.0
> > ```
> 
> ack! `protobuf` has had conflicting version bumps for a while now... it wouldn't be the worse thing to pin these in setup.cfg. We'll just have to remember to bump them later-- O(months)

Given that the bot is pretty smart (it knows about pinned dependencies anywhere in the dependency chain), and also checks on every commit, does `suru check` add any value?

dsschult:
Note: in many of our projects, we've started doing
```
pip install .[tests]
```
in the pytest action, and listing all the test deps in `setup.cfg`.

This has the advantage that the bot will keep the resulting requirements.txt file up to date, if it is ever needed (it's usually more of a record of what versions passed the CI tests).

file_catalog/server.py
```diff
@@ -1,51 +1,61 @@
...
+FC_AUTH_PREFIX = "resource_access.file-catalog.roles"
+FC_AUTH_ROLES = ["system"]
+
 
 # --------------------------------------------------------------------------------------
-# Utils
+# Auth
 # --------------------------------------------------------------------------------------
 
+if bool(os.environ.get('CI_TEST_ENV', False)):
```
ric-evans, line 38: This will be true for all values of `CI_TEST_ENV`. You may want to use `wipac-dev-tools`' `strtobool()`

file_catalog/server.py
```diff
@@ -99,225 +94,207 @@ def set_last_modification_date(metadata: types.Metadata) -> None:
...
+        "auth": {
+            "audience": config["AUTH_AUDIENCE"],
+            "openid_url": config["AUTH_OPENID_URL"],
+        },
+        "base_url": "/api",
+        "config": config,
+        "db": mongo,
+        "debug": debug,
+    }
+    if bool(os.environ.get('CI_TEST_ENV', False)):
```
ric-evans, line 129: see above comment

file_catalog/server.py
```diff
@@ -99,225 +94,207 @@ def set_last_modification_date(metadata: types.Metadata) -> None:
...
+        del args_dict["auth"]
+    args = RestHandlerSetup(args_dict)  # type: ignore
+    args["config"] = config
+    args["db"] = mongo
+
+    cookie_secret = ''.join(random.choice('0123456789abcdef') for _ in range(64))
```
ric-evans, line 135: I see what you're doing. It's clever certainly, but something like `hex(random_int)` maybe be clearer.

file_catalog/server.py
```diff
@@ -490,50 +464,37 @@ def initialize(self, **kwargs: Any) -> None:  # type: ignore[override]  # pylint
...
-            db_file = await self.db.get_file({'uuid': uuid})
-
-            if db_file:
-                db_file['_links'] = {
-                    'self': {'href': os.path.join(self.files_url, uuid)},
-                    'parent': {'href': self.files_url},
-                }
+        db_file = await self.db.get_file({'uuid': uuid})
+        if not db_file:
+            self.send_error(404, reason='File uuid not found')
```
ric-evans, line 472: You've already done a lot, so consider this optional. The `raise HTTPError`-syntax is easier to deal with all around.

file_catalog/server.py
```diff
@@ -627,20 +583,15 @@ def initialize(self, **kwargs: Any) -> None:  # type: ignore[override]  # pylint
...
     async def post(self, uuid: str) -> None:
         """Handle POST request.
 
         Remove location from the record identified by the provided UUID,
         and potentially the entire record.
         """
         # try to load the record from the file catalog by UUID
-        try:
-            db_file = await self.db.get_file({'uuid': uuid})
-        except pymongo.errors.InvalidId:
```
ric-evans, line 641: Is this no longer a concern?

file_catalog/server.py
```diff
@@ -713,21 +664,14 @@ def initialize(self, **kwargs: Any) -> None:  # type: ignore[override]  # pylint
...
+    @fc_auth(prefix=FC_AUTH_PREFIX, roles=FC_AUTH_ROLES)
     async def post(self, uuid: str) -> None:
         """Handle POST request.
 
         Add location(s) to the record identified by the provided UUID.
         """
         # try to load the record from the file catalog by UUID
-        try:
-            db_file = await self.db.get_file({'uuid': uuid})
-        except pymongo.errors.InvalidId:
```
ric-evans, line 726: see above comment

requirements-dev.txt
```diff
@@ -1,5 +1,10 @@
...
+crawler==0.0.2
+flake8==6.0.0
+more-itertools==9.0.0
+mypy==1.0.0
+pytest==7.2.1
+pytest-asyncio==0.20.3
+pytest-cov==4.0.0
+types-PyMySQL==1.0.19.2
+types-python-dateutil==2.8.19.6
+types-requests==2.28.11.12
```
ric-evans, line 10: you can throw this into setup.cfg to keep this up to date

resources/check_requirements.py
```diff
@@ -0,0 +1,41 @@
+# check_requirements.py
```
ric-evans, line 1: `wipac-dev-py-setup-action` does exactly this, via `pip-compile`

setupenv.sh
```diff
@@ -1,7 +1,7 @@
...
+#!/usr/bin/env bash
 unset PYTHONPATH
-virtualenv -p python3 env
+python3 -m venv env
 echo "unset PYTHONPATH" >> env/bin/activate
-. env/bin/activate
+source env/bin/activate
 pip install --upgrade pip
-pip install .
+pip install -r requirements.txt -r requirements-dev.txt
```
ric-evans, line 7: since you're packaging the FC. Why not do a `pip install .[dev]`?

tests/conftest.py
```diff
@@ -0,0 +1,116 @@
...
+import pytest
+import pytest_asyncio
+from rest_tools.client import RestClient
+
+from file_catalog.mongo import Mongo
+from file_catalog.server import create
+
+logger = logging.getLogger(__name__)
+
+MonkeyPatch = pytest.MonkeyPatch
```
ric-evans, line 26: Seems a bit odd this isn't part of an `import ... as ...` statement, but it works

tests/conftest.py
```diff
@@ -0,0 +1,116 @@
...
+        # provide the client as the fixture
+        yield fc_mongo
+
+    # -------------------------------------------------------------------------
+    # NOTE: *Your Unit Test Function Runs Here*
+    # -------------------------------------------------------------------------
+
+    # teardown_function
+    finally:
+        fc_mongo.close_me.close()
```
ric-evans, line 70: I didn't catch this earlier. Would a contxetmanager pattern be doable for this instance variable?

ric-evans (changes requested):
I think this wins largest PR of the year so far :smile:

file_catalog/server.py
dsschult: I'm personally fine with this, since you shouldn't define the env variable at all in prod.  But yeah, the conversion from str to bool is technically wrong (and unnecessary, since it automatically happens in the conditional).

file_catalog/server.py
dsschult: Actually, the new `secrets` module makes this much easier:
```python
cookie_secret = secrets.token_hex(16)
```

file_catalog/server.py
dsschult: I think it will never raise that error, since `InvalidId` is only for the `_id` field.

blinkdog:
> > > The bot updated requirements.txt and pinned the conflicted requirements to specific versions. `suru check` may suggest upgrades that can't really be done.
> > > ```
> > > googleapis-common-protos can be upgraded from 1.56.2 to 1.58.0
> > > protobuf can be upgraded from 3.20.3 to 4.21.12
> > > pyjwt can be upgraded from 2.5.0 to 2.6.0
> > > ```
> > 
> > 
> > ack! `protobuf` has had conflicting version bumps for a while now... it wouldn't be the worse thing to pin these in setup.cfg. We'll just have to remember to bump them later-- O(months)
> 
> Given that the bot is pretty smart (it knows about pinned dependencies anywhere in the dependency chain), and also checks on every commit, does `suru check` add any value?

It's a helper script; it just helps me keep tabs on things. My comment was more of a note-to-self: you can't upgrade these things quite yet.

file_catalog/server.py
blinkdog: Yeah, `pymongo.errors.InvalidId` no longer exists. (I think I had to go back to pymongo v1 to find it.)
So it doesn't really make sense to catch it any more.

There is a BSON version of the error, but I think it's like David says, only when Mongo has some internal trouble with `_id`.

That `db` is the file catalog's internal MongoDB client, so if we really want to do something about Mongo errors, it's probably better handled at that level. 

The keycloak auth decorator includes `catch_error`. For now, letting an exception bubble up and cause an error should work fine.

blinkdog:
> Note: in many of our projects, we've started doing
> 
> ```
> pip install .[tests]
> ```
> 
> in the pytest action, and listing all the test deps in `setup.cfg`.
> 
> This has the advantage that the bot will keep the resulting requirements.txt file up to date, if it is ever needed (it's usually more of a record of what versions passed the CI tests).

I've added this as an `[options.extras_require]` section in `setup.cfg` and modified the CI and setup scripts.
Does that work?

tests/conftest.py
blinkdog: Yeah, some time ago we used LGTM as a code quality scan.

For LTA it identified doing things like:
`import pytest`
`from pytest import MonkeyPatch`

The recommended workaround was just this; `MonkeyPatch = pytest.MonkeyPatch`

Personally, I like it better the import way, so I'll change it. Sorry, LGTM.

file_catalog/server.py
blinkdog: Good catch; I changed this to a simple "if present" check, no bool'ing.

file_catalog/server.py
blinkdog: Excellent. I like this new `secrets` module.

blinkdog:
I think we got most of the issues cleaned up, except #111 .
I'm going to cut bait and work on that tomorrow.

file_catalog/server.py
ric-evans: I'll take it :smile:

tests/conftest.py
ric-evans: > The recommended workaround was just this

tests/conftest.py
```diff
@@ -57,17 +56,15 @@ async def mongo() -> AsyncGenerator[Mongo, None]:
...
+    yield fc_mongo
 
     # -------------------------------------------------------------------------
     # NOTE: *Your Unit Test Function Runs Here*
     # -------------------------------------------------------------------------
 
     # teardown_function
-    finally:
-        fc_mongo.close_me.close()
+    fc_mongo.close_me.close()
```
ric-evans, line 68: I think this change is fine because a `fixture` uis technically a context manager. I'm not entirely sure `fc_mongo.close_me.close()` will execute if the consumer (unit test) raises an error. However, I'm not sure it matters in this case

ric-evans:
> > Note: in many of our projects, we've started doing
> > ```
> > pip install .[tests]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in the pytest action, and listing all the test deps in `setup.cfg`.
> > This has the advantage that the bot will keep the resulting requirements.txt file up to date, if it is ever needed (it's usually more of a record of what versions passed the CI tests).
> 
> I've added this as an `[options.extras_require]` section in `setup.cfg` and modified the CI and setup scripts. Does that work?

That's it! It's a smart bot :robot:

ric-evans (approved):

tests/conftest.py
blinkdog: I tested it. If the unit test raises an Exception (and that's all my test did), pytest's test runner catches it, then comes back to tear down the fixture, so it gets closed. That made me realize the try-finally block wasn't really necessary.

ric-evans:
Nice add! Let's see if this works:
closes #111

ric-evans:
What's the status of this PR?

blinkdog:
> What's the status of this PR?

Ready to merge. I'll probably do so this afternoon.

PR #146, opened 2023-12-21
add routestats env variables

dsschult:
Add routestats env variables so we can set window limit and overload time.

### DIFF
--- a/file_catalog/server.py
+++ b/file_catalog/server.py
@@ -125,6 +125,10 @@ def create(config: Dict[str, Any],
         "config": config,
         "db": mongo,
         "debug": debug,
+        "routestats": {
+            "window_time": config["ROUTESTATS_WINDOW_SIZE"],
+            "timeout": config["ROUTESTATS_OVERLOAD_LIMIT"],
+        },
     }
     if 'CI_TEST_ENV' in os.environ:
         del handler_setup["auth"]
--- a/file_catalog/server.py
+++ b/file_catalog/server.py
@@ -126,8 +126,9 @@ def create(config: Dict[str, Any],
         "db": mongo,
         "debug": debug,
         "route_stats": {
-            "window_time": config["ROUTESTATS_WINDOW_SIZE"],
-            "timeout": config["ROUTESTATS_OVERLOAD_LIMIT"],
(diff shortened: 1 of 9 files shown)


file_catalog/server.py
```diff
@@ -125,6 +125,10 @@ def create(config: Dict[str, Any],
         "config": config,
         "db": mongo,
         "debug": debug,
+        "routestats": {
```
blinkdog, line 128: I think it might need to be `route_stats` in the config object?

`rest_tools/server/handler.py:        route_stats = defaultdict(functools.partial(RouteStats, **config['route_stats']))`

https://github.com/WIPACrepo/rest-tools/blob/master/rest_tools/server/handler.py#L78

blinkdog (changes requested):
I'm not sure where RouteStats picks up the OVERLOAD_LIMIT; I'm still poking at rest-tools

file_catalog/server.py
dsschult: Sadly, you're right.  I must not have looked close enough.

file_catalog/server.py
```diff
@@ -125,6 +125,10 @@ def create(config: Dict[str, Any],
         "config": config,
         "db": mongo,
         "debug": debug,
+        "route_stats": {
+            "window_time": config["ROUTESTATS_WINDOW_SIZE"],
```
blinkdog, line 129: I'll lobby that you change the env var to `ROUTESTATS_WINDOW_TIME` because the RouteStats object has both a `window_size` and `window_time` field, and this is supplying a value for `window_time` while being named `ROUTESTATS_WINDOW_SIZE`, could be confusing.

I didn't realize at first that OVERLOAD_LIMIT was going to `timeout`; this one might be renamed `ROUTESTATS_TIMEOUT` instead, but this is just a suggestion.

blinkdog (changes requested):
Just one (okay two) more thing(s)...

file_catalog/server.py
dsschult: Sure, I can do those.

blinkdog (approved):
LGTM

